PR #43, opened 2015-01-14
Add .tarifa.json file and minimal settings in default ios configuration

peutetre:


tarifa.json
```diff
@@ -12,7 +12,11 @@
       "default": {
         "id": "org.lichess.mobileapp.dev",
         "product_name": "lichess dev",
-        "product_file_name": "lichess"
+        "product_file_name": "lichess",
+        "API_END_POINT": "http://en.lichess.org",
+        "SOCKET_END_POINT": "socket.en.lichess.org",
+        "GA_ID": "xxxxxxxxxxx",
```
veloce: I use `@ifdef` in preprocess to skip GA_ID on purpose, so this should be omitted in default conf.

tarifa.json
```diff
@@ -12,7 +12,11 @@
...
+        "version": "1.0.2"
```
veloce: Same for version: we don't need it in dev, so let's omit it

peutetre:
change done

veloce:
thanks!

PR #133, opened 2015-03-19
Share PGN for ai/otb modes

frankcalise:


veloce:
Thanks for this. I won't merge it as is because we'll use indeed the SocialMedia plugin.

I already installed it in this branch: https://github.com/veloce/lichobile/commits/urlSchemes

Can you uninstall clipboard plugin, merge `urlSchemes` in your branch and use `share` function instead?

Maybe you can share the pgn as text, should be easy.

frankcalise:
This is updated now, I am only able to test on Android, but it correctly prompts me for which app I'd like to share out to via text.

I just want to clean up the share icon (using Android one right now) depending on the platform, is there a utility to tell iOS versus Android at runtime already?

veloce:
You can use `window.cordova.platformId`. It can be either `ios` or `android`.

frankcalise:
Alright, fa-share is used for iOS and fa-share-alt for Android via the socialsharing plugin.

I'm new to most of these js libraries, I see mithril is being used, but couldn't come up with a better way to switch out the fa class based on platform. Let me know if the solution is acceptable in my change, 333e8e34397bdc031964a059ced26871e66ebac8 - I referenced [otb\promotion.js](https://github.com/veloce/lichobile/blob/1.4.x/project/src/js/ui/otb/promotion.js#L55-L57) so it seemed ok.

veloce:
It's ok but usually when we conditionally set classes we'd write something like this:
```javascript
m('div.myClassAlwaysHere', {
  className: (window.cordova.platformId === 'android') ? 'fa-share-alt' : 'fa-share'
}, ...);
```

Also as a side note, I'm ok to put specific platform icons where it's easy to do but I'm not keen on going to far in this direction since it would make our life harder to maintain the app. This is an hybrid app, with compromises for integration in native platforms.

frankcalise:
Ah, that is the syntax I was trying to work with. Thanks, I'll update it.

On the hybrid app note, point taken and understood. I mean you guys are the owners of the app, I'm just trying to pick a few pieces off the issues list and get some experience - feel free to direct me in any which way you feel is necessary :)

veloce:
I tested and it works well, so I'm merging this. Thank you!
I will modify a bit what you did when I'll merge the branch overlayRedesign though (we won't show PGN anymore but only share it).

Issue #20, opened 2015-01-13
online friends


ornicar:

frankcalise:
Is there an API available to retrieve a player's friends?

ornicar:
Yes. https://github.com/ornicar/lila/blob/master/doc/mobile/friends.md

It's a websocket API, because that data is by definition volatile.

veloce:
@frankcalise I'll be on holidays the next two weeks so I won't be able to help you much, but if you're looking to another feature to implement I suggest you #99 (add the clock option to otb games)
It's not API dependant so you can start right away, and it's quite interesting I think ;)

veloce:
Done

PR #142, opened 2015-03-29
Online friends feature

frankcalise:
Implements #20


project/src/js/i18n.js
```diff
@@ -23,6 +23,7 @@ var untranslated = {
   allowAnalytics: 'Allow anonymous statistics',
   shareGame: 'Share game',
   sharePGN: 'Share PGN',
+  onlineFriends: 'Online friends',
```
veloce: you don't want to add this: the `onlineFriends` key already exist and is translated. Here is only for untranslated keys.

project/src/js/socket.js
```diff
@@ -52,6 +52,23 @@ function createLobbySocket(lobbyVersion, onOpen, handlers) {
   return socketInstance;
 }
 
+function createFriendsSocket(friendsVersion, onOpen, handlers) {
```
veloce: Since it can be a multipurpose socket, I suggest a more generic name: `createSocket(name, onOpen, handlers)`

`version` here is not an api version, rather a event version. For now I guess you can just omit the parameter and pass `0` to StrongSocket.

project/src/js/ui/friends/index.js
```diff
@@ -0,0 +1,119 @@
...
+
+    var refreshList = function() {
+      // TODO sort the onlineFriends array
+
+      // update view
+      m.redraw();
+    }
+
+    xhr.friends().then(function(data) {
+      friendsSocket = socket.connectFriends('v1', requestFriends, {
```
veloce: same remark as above : version param is not supposed to be api version. just remove it for now.

veloce:
This is good. It's a basis for future challenge and user profile features. 
I can't merge it 'as is' though. At least as long as the online friends is merely a list. (Or removing the menu entry).
I think what we need first is to complete a debut version of the 'user profile' page. 
If you want to do it I guess we can start with basic infos: followers, following, location, language, member since, etc. And detailed ratings.

frankcalise:
Thanks for the feedback. I'll make the recommended changes from above.

I agree about the challenge and profile features. I had started to implement them but then wanted this pull request to just be about the online friends before going any further. Should those changes exist here? Or do I just create a new branch from here implementing those features?

As an initial direction, should I mimic the slide out menu's profile view when viewing another user's profile (not in a slide out, but when touching a list item from online friends)?

I already set up a /user/ route here and the ability to pull down the user profile json from the API.

veloce:
Challenges are more complicated so it's better in a separate PR. But user profile can be done in this PR I think. You can reuse the view of ratings in sliding menu yes.

veloce:
Merged in 1.5.x

PR #233, opened 2015-12-14
remove faulty PGN tag - fixes #232

### LINKED ISSUE #232: generating bad PGN?
https://www.reddit.com/r/chess/comments/3wqd1m/how_to_get_computer_analysis/

ornicar:
See https://github.com/veloce/lichobile/issues/232


Unihedro:
Instead of removing the tag wouldn't `[FEN ""]` also work to denote the default starting position?

veloce:
Well, I'll merge this. And if FEN is wanted by players I should have a mini spec to handle cases where FEN is faulty.

ornicar:
The FEN tag can be omitted for the standard initial position. Variants can include FEN to describe their initial position.

Issue #249, opened 2015-12-27
Add variant stats


veloce:

ornicar:
curl -H "Accept:application/vnd.lichess.v1+json" http://en.lichess.org/@/thibault/perf/blitz

freefal67:
I have submitted a pull request (https://github.com/veloce/lichobile/pull/258) adding this feature.

PR #258, opened 2016-01-20
Variant Statistics

freefal67:
Vincent,

As discussed, see below my addition of variant statistics to the lichobile app. Happy to jump on IRC or just chat over email if you'd like to discuss.

This feature is meant to address the issue raised here (https://github.com/veloce/lichobile/issues/249)

Regards,
Mark


ornicar:
can I see a screenshot?

veloce:
Great, thanks for this. I'll try to test it today.

freefal67:
See some screenshots here. 
http://m.imgur.com/a/BxBSJ

ornicar:
That's very nice, congrats. In my opinion the worst losses can be removed. Who wants to see that anyway?

freefal67:
Thanks Thibault.

I actually just noticed that the variant stats page is broken for variants
with little data. I probably need to avoid creating a link if there are no
games and do a bit more testing for the existence of certain elements so
variants with few games display correctly. The couple profiles I tested
this out on originally had lots of games so I didn't notice.

Agree on killing the worst losses (this exercise reminded me that I once
lost to someone rated below 900 in bullet). I'll do that as well and submit
a new pull request.

freefal67:
Ok, just made the changes in my prior comment.

veloce:
That's awesome, really. I'll merge this as soon as have the time.

I just looked at the code and all seems fine. Except that I saw you used IScroll in the view. Actually I'd prefer not to have IScroll, rather native scroll. I use IScroll only for user games view since it's an infinite scroll refreshed with ajax and I had perf issues when I tried with native scroll.

Every other pages use native scroll. It's very easy to setup. You can see how it's done for instance here: 

https://github.com/veloce/lichobile/blob/master/project/src/js/ui/settings/pieceThemes.js

Normally all you have to do is to add the `native_scroller` class to the scroll container.

freefal67:
I was importing IScroll in variantPerfCtrl.js, but am actually using native_scroller in variantPerfView.js.
I just committed a change that removed the reference. Let me know if I'm missing something bigger.

veloce:
Okay, I've tested it. I have just a few remarks that I'll put here (and some other in the code review):

* IMO the blank line (sum up line) is weird and useless, I'd remove it.
* the header title should be tweaked for small screens: instead of `playerName FullPerfName Stats` I'd have rather `playerName shortPerfName stats`. That is, add a 3rd field (short name) in perf type array here https://github.com/veloce/lichobile/blob/master/project/src/js/lichess/perfs.js and use it for that screen (uncapitalized)

project/src/js/ui/shared/perf.js
```diff
@@ -3,10 +3,14 @@ import { provisionalDeviation } from '../../lichess/perfs';
...
 
-module.exports = function(key, name, perf) {
-  return m('div.perf', {
-    'data-icon': gameIcon(key)
-  }, [
+module.exports = function(key, name, perf, user) {
+  var options = {'class': 'perf', 'data-icon': gameIcon(key)};
+  if (variantPerfAvailable(key, perf)) {
+    options['class'] += ' nav';
+    options.config = helper.ontouchY(goToVariantPerf(user, name));
```
veloce: I'd rather use `key` than `name` here.

project/src/js/ui/shared/perf.js
```diff
@@ -17,3 +21,10 @@ module.exports = function(key, name, perf) {
   ]);
 };
 
+function goToVariantPerf (user, name) {
+  return (() => m.route(`/@/${user.id}/${name}/perf`));
```
veloce: Same remark: `${key}` instead of `${name}`

project/src/js/ui/user/userXhr.js
```diff
@@ -34,3 +34,7 @@ export function games(userId, filter='all', page=1, feedback=false) {
 export function tv(userId) {
   return request(`/@/${userId}/tv`);
 }
+
+export function variantperf(userId, variant) {
```
veloce: I believe it should be variant key here. Instead of `variant.toLowerCase()`

project/www/i18n/en.json.new
```diff
@@ -0,0 +1 @@
...xitVariation":"enter/exit variation","keyYouCanDrawArrowsCirclesAndScrollToMove":"Press shift+click or right-click to draw circles and arrows on the board. You can also scroll over the board to move in the game.","newTournament":"New tournament","tournamentHomeTitle":"Chess tournament featuring various time controls and variants","tournamentHomeDescription":"Play fast-paced chess tournaments! Join an official scheduled tournament, or create your own. Bullet, Blitz, Classical, Chess960, King of the Hill, Threecheck, and more options available for endless chess fun.","tournamentNotFound":"Tournament not found","tournamentDoesNotExist":"This tournament does not exist.","tournamentMayHaveBeenCanceled":"It may have been canceled, if all players left it before the tournament starts.","returnToTournamentsHomepage":"Return to tournaments homepage","monthlyPerfTypeRatingDistribution":"Monthly %s rating distribution","nbPerfTypePlayersThisMonth":"%s %s players this month.","yourPerfTypeRatingisRating":"Your %s rating is %s.","youAreBetterThanPercentOfPerfTypePlayers":"You are better than %s of %s players.","youDoNotHaveAnEstablishedPerfTypeRating":"You do not have an established %s rating."}
```
veloce: I guess this file has been committed by mistake.

freefal67:
Ok, as stated in the commit message I did the following:
1) killed the sum line
2) deleted the extra 'env.json.new' file that was added in error
3) used 'key' rather than 'name' in the couple locations that you highlighted
4) created and used a shorter name for the variants in the header of variantPerfView

Thanks for the feeback and let me know if that all looks ok to you.

As an aside, I noticed that whereas earlier it didn't look like there were any merge conflicts, it appears that there now are. Note sure what they would be.

veloce:
Well done, thanks a lot!

ornicar:
:+1:

PR #281, opened 2016-02-18
Tournaments

freefal:
Vincent,

As discussed, I'm sending you a pull request for the tournament observation functionality.


ornicar:
:+1:

project/src/js/ui/tournament/list/tournamentListView.js
```diff
@@ -0,0 +1,76 @@
...
+    id: 'started',
+    label: 'In Progress'
+}, {
+    id: 'created',
+    label: 'Upcoming'
+}, {
+    id: 'finished',
+    label: 'Completed'
+}];
+
```
veloce: Here prefer this syntax:

```
function tabNavigation(currentTabFn) {
   return ...
}
```

Using arrow function is good for anonymous functions and expressions but not to declare a function.

project/src/js/ui/tournament/tournamentCtrl.js
```diff
@@ -0,0 +1,51 @@
+import socket from '../../socket';
+import * as utils from '../../utils';
+import * as xhr from './tournamentXhr';
+import socketHandler from './socketHandler';
+import helper from '../helper';
+import m from 'mithril';
+
+const tournament = m.prop([]);
```
veloce: `tounament` here should be a private field in function `controller` instead of defined globally.

project/src/js/ui/tournament/tournamentCtrl.js
```diff
@@ -0,0 +1,51 @@
...
+      if(clockInterval)
+        clearInterval(clockInterval);
+    }
+  };
+
+  socket.createTournament(tournament().socketVersion, id, socketHandler(returnVal));
+  return returnVal;
+}
+
+function reload (data) {
```
veloce: Same thing here: `reload()` and `tick()` should be functions defined inside controller. They are automatically private functions if you don't return them from `controller` function.

project/src/js/ui/tournament/tournamentCtrl.js
```diff
@@ -0,0 +1,51 @@
...
+  helper.analyticsTrackView('Tournament ' + id);
+
+  let clockInterval = null;
+  xhr.tournament(id).then(data => {
+    tournament(data);
+    clockInterval = setInterval(tick, 1000);
+    return data;
+  }, err => utils.handleXhrError(err));
+
+  let returnVal = {
```
veloce: Here don't use an intermediary `let`: use ```return { ... }'``` immediately.

project/src/js/ui/tournament/tournamentView.js
```diff
@@ -0,0 +1,156 @@
...
+  if(variant === 'standard') {
+    if(data.schedule)
+      variant = data.schedule.speed;
+    else if(data.position)
+      variant = data.position.name;
+    else
+      variant = '';
+  }
+
+  while(variant.length > 12) {
```
veloce: Here I'm not sure to understand what is done just at looking at the code. Generally prefer using functionnal style (`map`, `filter`, etc) and no `for` or `while` loops. 
Its' important for readability, maintainability and to ensure consistency within the project.

project/src/js/ui/tournament/tournamentView.js
```diff
@@ -0,0 +1,156 @@
...
+  let secs = time % 60;
+  if (hours > 0)
+    timeStr = preceedingText + ' ' + hours + ':' + pad(mins, 2) + ':' + pad(secs, 2);
+  else
+    timeStr = preceedingText + ' ' + mins + ':' + pad(secs, 2);
+  return timeStr;
+}
+
+function tournamentLeaderboard(data, trophies) {
+  let leaderboardData = data.standing.players.map(renderLeaderboardItem, trophies);
```
veloce: useless `let`: just write that expression in the template because the expression is short. For longer expressions it is ok to do so.

veloce: Here you use wrongly `array.map()`; the second parameter of `Array.map` is the `thisArg` (ie: the `this` value of the callback function)
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/map

Here I don't know what `trophies` is. But it's certainly not the `this` arg of `renderLeaderboardItem`. Usually you never need to explicitely bind a `this` when using map.
So just write `data.standing.players.map(renderLeaderboardItem)`

project/src/js/ui/tournament/list/tournamentListView.js
freefal: Borrowed this style from a polythene example. Agree your suggested syntax is better. Changed.

project/src/js/ui/tournament/tournamentCtrl.js
freefal: Agree. Done.

project/src/js/ui/tournament/tournamentCtrl.js
freefal: Agreed. Done.

project/src/js/ui/tournament/tournamentCtrl.js
freefal: The goal here is to be able to provide that return value to the socket before returning it. Does that make sense or do you prefer some other approach?

project/src/js/ui/tournament/tournamentView.js
freefal: Thanks. I ended up just simplifying it to taking the first word, which should cut off long problematic names like "Halloween Gambit".

project/src/js/ui/tournament/tournamentView.js
freefal: Yes, I was using 'this' in a very perverse was to carry in a boolean value. Now done more elegantly with a function creator that creates a closure for the boolean value.

.eslintrc
```diff
@@ -2,7 +2,6 @@
   "parser": "babel-eslint",
   "rules": {
     "quotes": [ 2, "single" ],
-    "linebreak-style": [ 2, "unix" ],
```
veloce: I don't care for that rule as long as you configure your git correctly for line endings:
https://help.github.com/articles/dealing-with-line-endings/#platform-windows

project/src/js/ui/tournament/tournamentView.js
```diff
@@ -0,0 +1,152 @@
...
+    timeStr = preceedingText + ' ' + mins + ':' + pad(secs, 2);
+  return timeStr;
+}
+
+function tournamentLeaderboard(data, showTrophies) {
+  return (
+    <div className='tournamentLeaderboard'>
+      <p className='tournamentTitle'> {i18n('leaderboard')} ({data.nbPlayers} Players)</p>
+      <table className='tournamentStandings'>
+        {data.standing.players.map(createLeaderboardItemRenderer(showTrophies))}
```
veloce: Wrong use of `Array.map` here. 

It will be better explained here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

But shortly, you should pass only one arg to the `map` function, and this is the `map` callback. The second arg of map is and object that will be bound to the `this` of the callback function. I won't explain here all the subtleties of `this` in javascript, but I'm pretty sure you don't want to use that feature here.

A regular `Array.map` should be just like so:

```js
my.array.map(function(el, index) {
    // do stuff with el which is the element being mapped over
    // optionnaly do stuff with the index of element, but this is much less frequent usage
}); 

If you use `map` correctly you should never have to increment vars in your `map` callback. So I won't write a comment on all those lines like below

```js
podiumRank++;
```

because it's all related to the wrong `map` usage. I'm pretty sure that when you will see the right solutions for this problem it will appear much more obvious and elegant to you :)

project/src/js/ui/tournament/tournamentView.js
```diff
@@ -0,0 +1,152 @@
...
+      </table>
+    </div>
+  );
+}
+
+function createLeaderboardItemRenderer(showTrophies) {
```
veloce: I think it should be even more simple here. Normally you don't need that factory. I guess you can directly write the `map` callback like so:

```js
renderLeaderboardItem(player, index) {
    // better to use a var to do this than to increment the index directly
    // it's less error prone and self explanatory
    const podiumRank = index + 1;
    
    ....
    return (
      ...
    );
}
```

.eslintrc
freefal: Sounds good. I mostly develop on linux, but sometime boot up a windows
machine, where I have git configured correctly.

On Fri, Feb 19, 2016 at 5:47 AM, Vincent Velociter <notifications@github.com

project/src/js/ui/tournament/tournamentView.js
freefal: Agree that "const podiumRank = index + 1;" is much more readable and have
made that change.



On Fri, Feb 19, 2016 at 6:10 AM, Vincent Velociter <notifications@github.com

freefal: I do think that I need the factory here as there is context ("Should the
callback display trophies?") that needs to be kept track of.

project/src/js/ui/tournament/list/tournamentListView.js
```diff
@@ -0,0 +1,76 @@
...
+        selectedTab: currentTabFn(),
+        activeSelected: true,
+        getState: (state) => {
+            currentTabFn(state.index);
+        }
+    }));
+}
+
+function tournamentListBody(ctrl) {
+  if (!ctrl.tournaments()) return null;
```
veloce: Here `ctrl.tournaments()` returns an array. Since an empty array evaluates to `true` you need to write this instead:

```js
if (ctrl.tournaments().length === 0) return null;
```

project/src/js/ui/tournament/tournamentView.js
veloce: Yep sure, you can keep it like that. It's perfectly ok.

project/src/js/ui/tournament/list/tournamentListView.js
freefal: ctrl.tournaments() returns an object with three properties ('started',
'created', 'finished'), each of which is an array, so I think this works as
written.
Check out curl -H "Accept:application/vnd.lichess.v1+json" "
http://en.lichess.org/tournament"

On Fri, Feb 19, 2016 at 8:16 AM, Vincent Velociter <notifications@github.com

veloce: I see.
The problem is  you declared tournaments like so:

```js
const tournaments = m.prop([]);
```

Until the promise is resolved, `ctrl.tournaments()` will return `[]`, which evaluates to `true`;

To fis this you need to declare tournaments like:

```js
const tournaments = m.prop({});
```

And then you need to check if object is empty, using lodash:

```js
if (isEmpty(ctrl.tournaments())) return null;
```

freefal: Done. I did the same for tournamentView as well.

On Fri, Feb 19, 2016 at 8:42 AM, Vincent Velociter <notifications@github.com

Issue #283, opened 2016-02-20
Tournaments TODO list


veloce:
- [x] variant icon in tournament header
- [x] mini board for featured game
- [x] podium
- [x] filter tournaments with supported variants
- [x] add ranking progress in leaderboard
- [x] add tournament player detail page (clickable from leaderboard)
- [x] implement possibility to play tournament
- [ ] tournament chat room
- [ ] tablet enhanced view
- [x] timeline entries
- [x] player rank in game screen 
- [x] remove chat during game
- [x] tournament countdown in game screen
- [x] berserk mode
- [ ] what about private tournaments?
- [x] browsable leaderboard

**Playable tournaments**
- [x] socket communication
- [x] return to tournament from round
- [x] tournament help page

**Player detail page**
- [x] scores sheet
- [x] stats
- [x] games list

veloce:
@freefal thank you very much for your first PR. You laid the first foundation, now here is the roadmap to finish properly the feature!

We don't have to do it all before a first release. I think that make them playable is a requirement though. I don't know if it's hard to do now we have already the 'view' code. I'm willing to help you.

Then feel free to send any PR concerning all the points listed above :) Now I'm working on analysis and there is a lot of work to do, so I won't touch tournament for now, that's why you can pick any of them if you want to!

veloce:
I'm not sure of what is to be done for playable tournaments, so the todo list is to be completed.

JoelMon:
@freefal, was it difficult? haha :)

freefal:
@veloce Ok, I have implemented:
1) "variant icon in tournament header"
2) "filter tournaments with supported variants" - I'm preventing people from joining crazyhouse tournaments, but letting them watch standings and games. Is there more to do here?
3) "add ranking progress in leaderboard"
4) "implement possibility to play tournament" - I didn't implement the "tournament help page"

To your list, I would add "Add a real podium for completed tournaments instead of just adding trophies to the leaderboard". I initially took the latter approach to save space, but the more I look at it, the more I dislike it.

I think that it is releasable at this stage, but defer to you.
I'll send you a PR tonight.

freefal:
@veloce I just sent you a PR with the featured game mini board and the podium.

freefal:
Next features on my list are chat and to make the leaderboard scrollable (flip through pages)
Don't think it matters whether they make it into the initial release of tournaments or not.

veloce:
[...]

PR #327, opened 2016-05-02
Playable Tournaments

freefal:
As I said in the issue discussion (https://github.com/veloce/lichobile/issues/283), tournaments are now playable and I've implemented some of the other enhancements we discussed.

There are still plenty of obvious enhancements to make the tournament feature better (and I will work on them over the coming weeks), but I believe it is releasable at this stage.

Please take a look and let me know what you think.


veloce:
Awesome. I'll review asap

project/src/js/ui/round/roundCtrl.js
```diff
@@ -404,7 +404,11 @@ export default function controller(cfg, onFeatured, onTVChannelChange, userTv, o
     if (this.chat) this.chat.onReload(rCfg.chat);
     if (this.data.tv) rCfg.tv = this.data.tv;
     if (this.data.userTV) rCfg.userTV = this.data.userTV;
+    let tournament = this.data.tournament;
+
     this.data = data(rCfg);
+
+    this.data.tournament = tournament;
```
veloce: Those two lines are mutually recursive declarations. This is not normal.

project/src/js/ui/round/view/button.js
```diff
@@ -298,5 +298,29 @@ export default {
...
+  },
+  returnToTournament: function(ctrl) {
+    function handler() {
+      ctrl.hideActions();
+      m.route('/tournament/' + ctrl.data.tournament);
+    }
+    return (
+      <button key="returnToTournament" config={helper.ontouch(handler)}>
+        <span className="fa fa-trophy" />
+        Return to Tournament
```
veloce: Here you should use `i18n('backToTournament')`

project/src/js/ui/round/view/button.js
```diff
@@ -298,5 +298,29 @@ export default {
...
+      </button>
+    );
+  },
+  withdrawFromTournament: function(ctrl) {
```
veloce: This button does not belong to game. I prefer to stick to what is done in the website. Thus the withdraw button should be in the tournament screen only. You can then take the i18n message `withdraw` which is then self explanatory.

project/src/js/ui/round/view/roundView.js
```diff
@@ -375,20 +375,30 @@ function renderGameEndedActions(ctrl) {
...
-    button.shareLink(ctrl),
-    button.sharePGN(ctrl),
-    button.newOpponent(ctrl),
-    button.answerOpponentRematch(ctrl),
-    button.cancelRematch(ctrl),
-    button.rematch(ctrl),
-    button.analysisBoard(ctrl)
-  ];
+  let buttons = null;
+  if (ctrl.data.tournament) {
```
veloce: What if you are spectator of a tournament? I think you should have also the regular spectator buttons.

Same for player actually: the newOpponent and all rematch related buttons should not be displayed in a tournament context ; it's not the same for the other buttons that should remain.

So the logic must be changed to take this into account.

project/src/js/ui/tournament/detail/socketHandler.js
```diff
@@ -0,0 +1,23 @@
+import * as xhr from '../tournamentXhr';
```
veloce: Why socketHandler file? in current master it is not needed.

project/src/js/ui/tournament/detail/socketHandler.js
```diff
@@ -0,0 +1,23 @@
...
+export default function(ctrl) {
+  let reload = throttle((c) => {
+    xhr.reload(c.tournament().id).then(c.reload);
+  }, 2000);
+
+  const handlers = {
+    reload: function() {
+      reload(ctrl);
+    },
+    resync: function() {
```
veloce: resync action appear to be not used anymore. (see lila/ui/tournament/ code)

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -1,56 +1,83 @@
...
 export default function controller() {
-  const id = m.route.param('id');
+	helper.analyticsTrackView('Tournament details');
 
-  helper.analyticsTrackView('Tournament details');
+	const tournament = m.prop();
+	const hasJoined = m.prop(false);
 
-  const tournament = m.prop();
+	function reload(data) {
```
veloce: Bad indent.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -1,56 +1,83 @@
...
 
-  function reload(data) {
-    tournament(data);
-    if (data.socketVersion) {
-      socket.setVersion(data.socketVersion);
-    }
-    m.redraw();
-  }
+		if (data.socketVersion) {
+			socket.setVersion(data.socketVersion);
```
veloce: Bad indent

project/src/js/ui/tournament/detail/tournamentDetailView.js
```diff
@@ -43,49 +57,80 @@ function tournamentContentFinished(data) {
...
+      ctrl.withdraw();
     else
-      variant = '';
+      ctrl.join();
   }
 
+  if (ctrl.tournament().variant === 'crazyhouse')
+    return ('');
+
+  return (m('.pe-dark-theme .joinButton',
```
veloce: I don't use polythene anymore. You should remove the dependency and not have any class name of polythene.

project/src/js/ui/tournament/detail/tournamentDetailView.js
```diff
@@ -43,49 +57,80 @@ function tournamentContentFinished(data) {
...
+  function buttonAction () {
+    if (ctrl.hasJoined())
+      ctrl.withdraw();
     else
-      variant = '';
+      ctrl.join();
   }
 
+  if (ctrl.tournament().variant === 'crazyhouse')
+    return ('');
```
veloce: Don't return a empty string. return `null` if nothing to display.

Instead of filtering the variant like that you should use the `availableVariants` array in settings.js

project/src/js/ui/tournament/detail/tournamentDetailView.js
```diff
@@ -94,6 +139,21 @@ function variantInfo(data) {
   return variant;
 }
 
+function variantKey(data) {
```
veloce: You should not have to deduce the variant key from the time. Actually this info you get it from the server. And it's not `variant` key but `perf` key which have the `blitz` or `classical`, etc values.

project/src/js/ui/tournament/tournamentXhr.js
```diff
@@ -9,5 +9,17 @@ export function tournament(id) {
 }
 
 export function reload(id) {
-  return request(`/tournament/${id}`, { background: true });
+  return request('/tournament/' + id, { background: true });
+}
+
+export function resync(id) {
```
veloce: you can't send hardcoded socketVersion. And `resync` is not useful anymore I believe. Check lila code or ask @ornicar

project/src/styl/tournament.styl
```diff
@@ -96,3 +106,10 @@ td.tournamentPoints
     border-bottom-color darken(bglight, 15%)
     &.active
       background-color darken(bglight, 5%)
+
+.joinButton
+  float: right
+
+.content .pe-dark-theme .pe-button--raised.pe-button .pe-button__content
```
veloce: polythene class names to remove.

project/src/js/ui/round/roundCtrl.js
freefal: I'm not sure that it's recursive. The idea here is to add a "tournament" variable to "this.data" so that at the end of a game we can reference it.

project/src/js/ui/round/view/button.js
freefal: Ok, will do.

project/src/js/ui/round/view/button.js
freefal: The game screen on the website has a "withdraw" button (http://imgur.com/Lzu5aWp). It's a nice convenience button to have so that the user doesn't have to hit the "back to tournament" button from the game page and then try to hit the "withdraw" button on the tournament page before the tournament sockets sends a redirect.

project/src/js/ui/round/view/roundView.js
freefal: Agreed. I will add back the other buttons.

project/src/js/ui/tournament/detail/socketHandler.js
freefal: Ok, I will ask @ornicar how I should now deal with the socket version and resync.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
freefal: Not sure what happened with the indenting. It looked fine to me in Atom and eslint didn't complain. I'll figure out what's happening and fix.

project/src/js/ui/tournament/detail/tournamentDetailView.js
freefal: Ok, I wasn't seeing a `perf` key in the tournament object. Will look again and if not there, I'll ask @ornicar if he can add.

project/src/js/ui/tournament/detail/tournamentDetailView.js
freefal: Ok, I am just using it for that join/withdraw button at this point. I will find another way to get a nice button.

project/src/js/ui/tournament/detail/tournamentDetailView.js
freefal: Ok, thanks. Wasn't aware of that array. I'll make the change.

project/src/js/ui/round/view/roundView.js
freefal: I'm a bit confused about how the current control flow for the `end:` handler within round's `socketHandler` works. It seems to prevent the popup with the game action from being created if the user is a spectator. Therefore I'm not sure how the button code for the spectators would ever be relevant. I'm sure that I'm missing something, but for the time being I've changed:

`if (!ctrl.data.player.spectator) {`

to read:
`if (ctrl.data.tournament || !ctrl.data.player.spectator) {`

so as not to affect things outside of tournaments.

project/src/js/ui/tournament/detail/socketHandler.js
freefal: Ok, removed socketHandler file and the resync handler and just put the remaining handlers in the tournamentDetailCtrl.js file

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
freefal: File looked fine for me locally, but I reindented anyway with atom. Hopefully that fixes the issue.

project/src/js/ui/round/view/button.js
veloce: Ok then let's keep it, but use the translated string. We should be careful to not add too much buttons since the popup size is limited though.
So maybe you did the right thing in the first place: share pgn and copy link should be removed on tournament.

project/src/js/ui/round/roundCtrl.js
veloce: I see now. ok.

project/src/js/ui/round/view/roundView.js
```diff
@@ -375,21 +375,43 @@ function renderGameEndedActions(ctrl) {
...
+        button.returnToTournament(ctrl),
+        button.shareLink(ctrl),
+        button.sharePGN(ctrl)
+      ];
+    }
+    else {
+      buttons = [
+        button.returnToTournament(ctrl),
+        button.withdrawFromTournament(ctrl),
+        button.shareLink(ctrl),
```
veloce: Maybe remove those two buttons if there are to many buttons.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -24,33 +26,66 @@ export default function controller() {
...
     }
-
     m.redraw();
   }
 
+  function join(id) {
+    if (!id) {
+      id = tournament().id;
+    }
+    xhr.join(id).then(() => {
```
veloce: `join` and `withdraw` xhr should be throttled too, at 1 second.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -24,33 +26,66 @@ export default function controller() {
...
+  }
+
+  function withdraw(id) {
+    if (!id) {
+      id = tournament().id;
+    }
+    xhr.withdraw(id).then(() => {
+      hasJoined(false);
+      m.redraw();
+    }, err => utils.handleXhrError(err));
```
veloce: Avoid using the `.then(successFunction, errorFunction)` form.
Prefer the notation:
```
stuff()
.then(sucess => ...)
.catch(error => ...)
```

which is more readable and a better practice when working with promises.

veloce: Here it could be rewritten as:
```js
xhr.withdraw(id)
.then(() => {
    ....
})
.catch(utils.handleXhrError);
```

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -24,33 +26,66 @@ export default function controller() {
...
+  }
+
+  let id = m.route.param('id');
+  let action = m.route.param('action');
```
veloce: Use `const` keyword here (actually try to use `const` all the time and use `let` or `var` only when necessary). And those two variables should be declared at the top of the controller to be coherent with the rest of the project.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -24,33 +26,66 @@ export default function controller() {
...
+  if (action && action === 'withdraw') {
```
veloce: I'm not sure having an `action` route parameter is the right way to do this. There is no such other behavior in the app. This is a client side app and url params should only have ids or params used to read data, not perform actions (contrary to what you would do server side).

The withdraw action is actually carried by the server. So your job is to send the xhr and update your business data after server response. Which is already what you do, so you just need to do it directly without the mithril route (this is not a good practice to do so in mithril I believe).

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -24,33 +26,66 @@ export default function controller() {
...
+    hasJoined,
+    join,
+    withdraw,
     reload,
     onunload: () => {
       socket.destroy();
-      if (clockInterval) {
-        clearInterval(clockInterval);
-      }
+      if (clockInterval)
```
veloce: Actually I prefer the bracket notation :)

project/src/js/ui/tournament/detail/tournamentDetailView.js
```diff
@@ -34,58 +47,84 @@ function tournamentBody(ctrl) {
...
+
+  function buttonAction () {
+    if (ctrl.hasJoined())
+      ctrl.withdraw();
     else
-      variant = '';
+      ctrl.join();
   }
 
+  if (ctrl.tournament().isFinished || settings.game.supportedVariants.indexOf(ctrl.tournament().variant) < 0) {
```
veloce: :+1:

project/src/js/ui/round/view/roundView.js
freefal: It actually ends up looking fine to me.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
freefal: Got it. Thanks.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
freefal: Ok, started doing this in a couple places. Will go through an make this change everywhere appropriate.

project/src/js/ui/round/roundCtrl.js
veloce: Actually it's better to be consistent with the lines above and write:
```
if (this.data.tournament) rCfg.tournament = this.data.tournament;
```

project/src/js/ui/round/socketHandler.js
```diff
@@ -63,7 +63,7 @@ export default function(ctrl, onFeatured, onUserTVRedirect) {
       if (ctrl.data.game.speed === 'correspondence') {
         removeOfflineGameData(ctrl.data.url.round.substr(1));
       }
-      if (!ctrl.data.player.spectator) {
+      if (ctrl.data.tournament || !ctrl.data.player.spectator) {
```
veloce: I'd remove this. Not useful to show menu if spectator of a tournament.

project/src/js/ui/tournament/detail/tournamentDetailCtrl.js
```diff
@@ -1,50 +1,79 @@
...
+  function withdraw(id) {
+    xhr.withdraw(id).then(() => {
+      hasJoined(false);
+      m.redraw();
+    }).catch(utils.handleXhrError);
+  }
+
+  const id = m.route.param('id');
+
+  const throttled_reload = throttle(() => {
```
veloce: don't use `snake_case` but `camelCase` for javascript identifiers.

project/src/js/ui/round/socketHandler.js
freefal: I have removed. I could see being able to return to the tournament with a
popup button being marginally useful, but acknowledge that there is a big
back button right there at the top of the page and don't feel strongly.

project/src/js/ui/round/roundCtrl.js
freefal: Much better. Made this change.

veloce:
Thanks for your work!

PR #336, opened 2016-05-23
Add Navigation Buttons to Leader Board

freefal:
Also add stripe to left of user's name in leader board.

### DIFF
--- a/project/src/js/ui/tournament/detail/tournamentDetailView.js
+++ b/project/src/js/ui/tournament/detail/tournamentDetailView.js
@@ -197,12 +197,12 @@ function tournamentLeaderboard(ctrl) {
       </table>
 
       <div className="navigationButtons">
-        <button className={'navigationButton' + (back ? '' : 'disabled')} data-icon='W' config={helper.ontouch(back ? ctrl.reload(data.id, 1) : null)} />
-        <button className={'navigationButton' + (back ? '' : 'disabled')} data-icon='Y' config={helper.ontouch(back ? ctrl.reload(data.id, page - 1) : null)} />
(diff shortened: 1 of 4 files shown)


project/src/js/ui/tournament/detail/tournamentDetailView.js
```diff
@@ -182,23 +182,48 @@ function timeInfo(time, preceedingText) {
...
+  const user = session.get();
+  const userName = user ? user.username : '';
+
   return (
     <div key="leaderboard" className='tournamentLeaderboard'>
       <p className='tournamentTitle'> {i18n('leaderboard')} ({data.nbPlayers} Players)</p>
+
       <table className='tournamentStandings'>
-        {data.standing.players.map(renderLeaderboardItem)}
+        {data.standing.players.map(createLeaderboardItemRenderer(userName))}
```
veloce: Instead of using a factory you can use [Function.prototype.bind](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Function/bind).

It's more readable, and javascript-like:

Ex:
```
data.standing.players.map(renderLeaderboardItem.bind(undefined, userName)

...

function renderLeaderboardItem(userName, player) {
     // do your stuff with the bound function
}

```

veloce:
:+1: 

Thank you very much! I haven't tested it but from the code perspective it's all good

project/src/js/ui/tournament/detail/tournamentDetailView.js
freefal: Thanks for the mini lesson on bind. Will use in the future where it makes sense. Committed this change and pushed it.

PR #339, opened 2016-05-26
fix typescript imports

vdebergue:
This will make the project compiles correctly with typescript. 

I haven't tested the application yet (it takes some time to configure android ^^)


veloce:
awesome man! :)

I can help you to build the app if you need.

veloce:
Or you can just test it with `gulp watch` and the browser.

vdebergue:
If you are @zen tomorrow we can have a look. I didn't manage to turn off chrome security on osx. Will need some more time to figure it out

veloce:
Ok let's do this. I'll add the security stuff in the readme because a lot of people have asked me about it.

PR #343, opened 2016-05-31
Player Modal on Tournament Leader Board

freefal:


project/src/js/ui/layout.js
```diff
@@ -67,7 +67,7 @@ export default {
         {challengeForm.view()}
         {friendsPopup.view()}
         {lobby.view()}
-        {overlay ? overlay() : null}
+        {overlay ? overlay.map((t) => t()) : null}
```
veloce: that is not necessary and is likely to break other views. 

Overlay should return an array of vdom, in the end it's the same as mapping over an array of functions returning a single vdom. But I'd like to not change layout api in fear to not break things

project/src/js/ui/tournament/detail/tournamentDetailView.js
```diff
@@ -16,17 +17,24 @@ export default function view(ctrl) {
...
+  const faqOverlay = renderFAQOverlay.bind(undefined, ctrl);
+  const playreInfoOverlay = renderPlayerInfoOverlay.bind(undefined, ctrl);
 
-  return layout.free(headerCtrl, body, footer, overlay);
+  return layout.free(headerCtrl, body, footer, [faqOverlay, playreInfoOverlay]);
 }
 
-function renderOverlay(ctrl) {
+function renderFAQOverlay(ctrl) {
   return [
```
veloce: so here:

```js
function renderOverlay(ctrl) {
  return [
     renderFAQOverlay(ctrl),
     playerInfo.view(ctrl.playerInfoCtrl)
  ];
}
```

project/src/js/ui/tournament/playerInfo.js
```diff
@@ -0,0 +1,150 @@
...
+import * as utils from '../../utils';
+import i18n from '../../i18n';
+
+export default {
+  controller: function(tournament) {
+    let isOpen = false;
+    const playerData = m.prop();
+
+    function open(p) {
+      xhr.playerInfo(tournament().id, p.name).then((data) => {
```
veloce: this is only stylistic matters, but I like to write .`then` and `.catch` in a new like as it's more readable I think.

```js
stuff()
.then(doSthAfter)
.catch(handleError);
```

veloce: Also when you have a single arg in arrow function you can omit the parenthesis:

```js
.then(data => {
   ....
})
```

PR #345, opened 2016-06-09
Upgrade stockfish.js to ddugovics Stockfish 7

niklasf:
Let's upgrade the JavaScript polyfill to [the one now used by lichess.org](https://github.com/ornicar/lila/blob/master/public/vendor/stockfish7.js). Advantages:

* Stockfish 7
* Variant support by ddugovic
* 1 MB instead of 3 MB


veloce:
Awesome! I need also to update the native part to benefit from the changes in three-check analysis.

PR #351, opened 2016-06-12
Switch to HTTPS

niklasf:
lichess.org is now using TLS \o/. The old API and socket endpoints will be available as long as we need them, but let's switch as soon as possible. This has been changed:

* http:// -> https://
* ws://socket.{lang}.lichess.org -> wss://socket.lichess.org
* Alternative ports 9025, 9026, 9027, 9028, 9029 now accept secure websockets

In this PR the protocol is now included in env.json, so you could still configure http:// for local testing. Every developer will have to update their env.json, though.

Tested:

* [X] Browser
* [x] Android (I can test later on 6.0.1)
* [ ] iPhone (Don't have it)

### DIFF
--- a/project/env.json.example
+++ b/project/env.json.example
@@ -1,4 +1,4 @@
 {
-  "API_END_POINT": "en.stage.lichess.org",
-  "SOCKET_END_POINT": "socket.en.stage.lichess.org"
+  "API_END_POINT": "https://en.stage.lichess.org",
+  "SOCKET_END_POINT": "wss://socket.stage.lichess.org"
 }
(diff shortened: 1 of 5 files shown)


veloce:
I've been waiting for this a looong time :) I'm not at home atm but I'll work on this asap, at the end of the week or next week!

niklasf:
\o/

I am curious: Do you know how long it takes until almost everyone has updated after a release?

veloce:
Ha. Well it's hard to say, but within too weeks almost everyone has updated I'd say. A few people still use very old versions but we can't do nothing against it.

Issue #254, opened 2016-01-16
Add standalone chess clock


JoelMon:
The _over the board_ mode is great when you want to play an impromptu game of chess with someone and you don't have a chessboard handy. I would love for a chess clock to be built into lichobile for those days where you have a chess board handy but no chess clock (or your chess clock's batteries die).

A standalone chess clock would compliment lichobile well.

ornicar:
I agree. For now I'm using chess.com clock app, it does the job very well.

quinncomendant:
It wasn't mentioned here, but it seems the over-the-board mode doesn't have any chess clock at all (even when “you don't have a chessboard handy”). Is there a separate feature ticket for that? My GF insists we time our games—I'm that slow. ;)

JoelMon:
That would be an excellent feature as well.

veloce:
That's planned too, maybe soon.

JoelMon:
Wonderful! Thank you for the clock @freefal :)

PR #354, opened 2016-06-14
Standalone Clock

freefal:
Add a standalone chess clock to be used with a real board.
Closes https://github.com/veloce/lichobile/issues/254


Unihedro:
There are a few clock modes that the main site doesn't have.

freefal:
Yes, the site just has Sudden Death and Increment, which is the right decision for online play. Here, I've additionally included Simple Delay, Bronstein Delay, Hour Glass, and Staged time controls.

project/src/js/ui/clock/clockView.js
```diff
@@ -0,0 +1,68 @@
+import h from '../helper';
+import layout from '../layout';
+import i18n from '../../i18n';
+import clockSettings from './clockSettings';
+import { formatTimeinSecs } from '../../utils';
+import sound from '../../sound';
+
+export default function view(ctrl) {
+  window.StatusBar.hide(); // Put this here instead of ctrl so it is reapplied after a phone lock and unlock
```
veloce: But here is it called on each redraw, which is not a good thing. You can simply listen to the cordova 'resume' event to hide again the status bar.

You'll need to subscribe to the event and not forget to unsuscribe on unload so the behavior don't apply to other screens.

freefal: Neat. I didn't know there was as resume callback. I will change.

On Tue, Jun 21, 2016 at 9:48 AM, Vincent Velociter <notifications@github.com

project/src/js/ui/clock/clockCtrl.js
```diff
@@ -0,0 +1,518 @@
...
+      window.plugins.insomnia.allowSleepAgain();
+      window.StatusBar.show();
+      if (clockObj().clockInterval) {
+        clearInterval(clockObj().clockInterval);
+      }
+    }
+  };
+}
+
+function simpleClock(time, draw) {
```
veloce: I think the name `draw` here (and every other occurence below) is ambiguous because it can be confused with a game draw, and in the whole app it is named `redraw`.

Actually the parameter is unnecessary, every `draw()` can be rewritten as `m.redraw()`. It would clarify things imo.

veloce:
That's awesome, really! Thank you very much for this. Apart from the two comments I made I can merge this.

I saw that you didn't used a monospace font for the clock. That can be an issue on some phone maybe, I don't know. Even if in the browser it renders well (I guess thanks to the flexbox layout).

I suggest to use Cousine as font-family so it's a monospace we know that will render the same on all the phones, and it's the same as the clock of online games of the app.
We just need to adjust a bit the letter-spacing when having long strings like '2 : 00 : 00'.

project/src/js/ui/clock/clockCtrl.js
freefal: My idea here was to keep the clock code relatively functional by using
callbacks rather than calling functions from upper scope (e.g.,
"m.redraw()") so that we can more easily use them generically in other
contexts (e.g., timed OTB games). Don't feel strongly though and we can
always revert if useful so will make this change.


On Tue, Jun 21, 2016 at 9:53 AM, Vincent Velociter <notifications@github.com

freefal:
I will try that font and see how it looks.

On Tue, Jun 21, 2016 at 9:57 AM, Vincent Velociter <notifications@github.com

veloce:
I noticed another problem: how do you get back to the app? On android there is the back button, but on iOS there is no such way to go back. Maybe a home button icon that brings back the the home page.

freefal:
Hmm...I only work on Android and Chrome so I hadn't noticed. Good point. I
just added a home button.

PR #369, opened 2016-07-02
add color select for game seek

madnight:
Hi i simply added color select like this:

since i'm new to your application: does it work this way?

Im not sure how to test properly but i tested it on the live system
<pre><code>window.lichess = {
        apiEndPoint: "https://en.lichess.org",
        socketEndPoint: "wss://socket.lichess.org",
        mode: "dev",
      };</code></pre>
and it works there : )

### DIFF
--- a/project/src/js/settings.js
+++ b/project/src/js/settings.js
@@ -231,7 +231,8 @@ export default {
       increment: localstorageprop('settings.game.human.increment', '0'),
       days: localstorageprop('settings.game.human.days', '2'),
       mode: localstorageprop('settings.game.human.mode', '0'),
-      membersOnly: localstorageprop('settings.game.human.membersOnly', false)
+      membersOnly: localstorageprop('settings.game.human.membersOnly', false),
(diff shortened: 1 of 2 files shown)


flugsio:
Color selection is not available for exotic variants in rated games, only casual. Antichess, Atomic, Horde, Racing Kings

madnight:
ok how about adding color selection only for the variant Standard?

ornicar:
I vote :-1:, random color is an excellent default value, and the UI is cluttered already.

madnight:
Well its really annoying that you cant choose the color in the app, i dont understand why the app should miss such a basic features. If im quite honest this missing features was the main reason why i coudn't use the app in many cases, because players want to try openings for black or white and therefore need to choose the color, otherwise people select random and then always abort if they get the wrong color wich is also very annoying. And this already there for Computer Play so i dont get the UI clutter argument, you basicly make the UI more consistent by adding this feature.

PR #370, opened 2016-07-03
update color select only for variant standard

madnight:
> Color selection is not available for exotic variants in rated games, only casual. Antichess, Atomic, Horde, Racing Kings

now color select is only available for Variant Standard and AI Games 
(this is also possible on the website)


veloce:
So I decided to allow to choose side for casual games. It's done here, because I thought it was better to refactor the forms: https://github.com/veloce/lichobile/commit/a99ef2c45f9b14b0c6d9d82a3032155c46359fe6

Thus I don't need you PR, thanks anyway for bringing the matter.

madnight:
thx thats all i ever wanted :+1:

PR #392, opened 2016-08-04
Ping and Server Lag Times

freefal:
This pull request adds ping and server lag times to the navigation menu.
Closes Issue 362: https://github.com/veloce/lichobile/issues/362


project/src/js/socket.js
```diff
@@ -20,13 +20,24 @@ let redrawOnDisconnectedTimeoutID;
...
+
 const defaultHandlers = {
   following_onlines: handleFollowingOnline,
   following_enters: name => autoredraw(() => friendsApi.add(name)),
   following_leaves: name => autoredraw(() => friendsApi.remove(name)),
   challenges: data => {
     challengesApi.set(data);
     m.redraw();
+  },
+  mlat: mlat => {
```
veloce: That can't be done like this. Since it will trigger a `redraw` on each ping event, that's a lot of useless redraws. The redraws should happend only when menu is open, and not all the time.

freefal: Great point. I will fix.

On Tue, Aug 16, 2016 at 7:55 AM, Vincent Velociter <notifications@github.com

freefal:
@veloce: I think this is now good to go.

veloce:
Yep it's in my todo list, I'll test it asap ;)

Issue #382, opened 2016-07-19
Over the Board Crazy House - Undo cant place piece from graveyard


Spooky-0:
The title basically says it all. If you play crazyhouse over the board, and missclick or need to Undo for whatever reason, you are unable to place a piece from the graveyard (A current workaround is to undo two moves, and then redo(no via > button, but actually playing the move again).

PR #397, opened 2016-08-10
Allow new drops in OTB even when replaying to facilitate drops after a takeback

freefal:
Closes https://github.com/veloce/lichobile/issues/382

### DIFF
--- a/project/src/js/ui/otb/otbCtrl.js
+++ b/project/src/js/ui/otb/otbCtrl.js
@@ -62,7 +62,7 @@ export default function controller() {
 
   const onUserNewPiece = function(role, key, meta) {
     const sit = this.replay.situation();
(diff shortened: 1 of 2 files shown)


veloce:
I can't accept it as is, since we do want to block drag'n drop from pockets while replaying in other modes than OTB.

Must use the `isOTB` boolean from crazyView.pocket to have that logic.

freefal:
Great. Thanks for the pointer. I will fix.

On Tue, Aug 16, 2016 at 7:42 AM, Vincent Velociter <notifications@github.com

veloce:
Fixed in https://github.com/veloce/lichobile/commit/88f6d60b067f63e15f8467362a1f9a8078eadae0

PR #400, opened 2016-08-12
Add 'keyboardMove' to preferences

freefal:
Currently no settings can be set because this setting is missing in the request.


ornicar:
I fixed it on server side. I'll make all new preferences optional from now on.

Issue #273, opened 2016-02-02
OTB - Improve board coordinates for black


JoelMon:
Hello,

When playing _over the board_ white can see the coordinates for the board just fine. When it's black's turn, the coordinates remain oriented for white. This is annoying for black since the letters and numbers are all upside down from their viewpoint making it hard if they would like to record the game or whatnot.

I propose an improvement for the coordinate markings on the board. Either:
- Have the coordinate system written down for both players the same way [tournament boards](http://www.houseofstaunton.com/media/catalog/product/cache/3/full/9df78eab33525d08d6e5fb8d27136e95/c/o/combopr-4_1_1_.jpg) usually are.
- Have the coordinate system _flip_ on each player's turn.

I think having the coordinates printed along the sides for each player is the way to go since folks usually write down their move after they moved their piece.

veloce:
Sure, I can do that.

Manuelp2:
Hi, I've worked on this issue and implemented a solution.
I just wanted to ask where would the preference gathering be best.

So in "View.js" line 174 at "project/node_modules/chessground-mobile/src/view.js"
I just need to make a conditional, pseudocode below

`if (ctrl.data.coordinates) {

+  if (OTB && symmetricalCoordinatesEnabled) {
// WHITE COORDINATES
    if (pos[1] === (asWhite ? 1 : 8)) attrs['data-coord-white-x'] = file;
    if (pos[0] === (asWhite ? 8 : 1)) attrs['data-coord-white-y'] = rank;
// BLACK COORDINATES
    if (pos[1] === (asWhite ? 8 : 1)) attrs['data-coord-black-x'] = file;
    if (pos[0] === (asWhite ? 1 : 8)) attrs['data-coord-black-y'] = rank;
} else { 
// Original code
    if (pos[1] === (asWhite ? 1 : 8)) attrs['data-coord-x'] = file;
    if (pos[0] === (asWhite ? 8 : 1)) attrs['data-coord-y'] = rank;
}`

depending on a preference stored in settings.otb.seeSymmetricCoordinates (new preference addition).

So the question is how do I access this preference from "View.js" according to your preferred workflow?

Manuelp2:
I'm almost finished.
The last thing I need is to reload the board afterwards.
In the `/project/src/js/ui/otb/otbView.js` Class.

@veloce Any pointers on how to do it? or could you do it for me?
Thanks.

veloce:
[...]

PR #403, opened 2016-08-17
Fix for issue #273

Manuelp2:
This is a fix for issue #273 
It requires another pull on the chessground repository (https://github.com/veloce/chessground-mobile/pull/1)

In order to not mess with the surrounding layout and keep maximum board space, coordinates are inside the squares.

ISSUE:
The board is not updated with `m.redraw` but getting out of OTB and in again does the trick.


veloce:
Sorry for the delay, I will now look at this.

veloce:
Thanks! I also fixed the redraw issue.

PR #441, opened 2016-10-07
chat: handle shadowban, deleted messages, anti-spam - fixes #431

### LINKED ISSUE #431: chat protection against shadowbanned users
The `/account/info` now has a `troll: true` field (along the `kid` field) telling if the current user is shadowbanned himself; in which case they can see other shadowbanned users messages.

The websites then accordingly filters chat messages like this: https://github.com/ornicar/lila/blob/master/ui/chat/src/discussion.js#L66-L77

Note that it also drops duplicated lines, as an additional flood protection, and handles ad spammers.

ornicar:
UNTESTED code, very probably it doesn't work.

### DIFF
--- a/project/src/js/session.ts
+++ b/project/src/js/session.ts
@@ -83,6 +83,10 @@ function isKidMode() {
   return session && session.kid;
 }
 
+function isShadowban() {
+  return session && session.troll;
+}
+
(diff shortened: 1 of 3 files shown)


veloce:
Seems to work ;)

Issue #424, opened 2016-09-11
Send bye msg when quitting online game


veloce:
https://github.com/ornicar/lila/issues/2237

ornicar:
bump!

PR #445, opened 2016-10-11
[WIP] Say 'bye' when leaving a game so other player can claim victory

freefal:
Closes https://github.com/veloce/lichobile/issues/424


veloce:
I think this is not enough: we should also handle the case when user puts app in background, or closes the app.

veloce:
@freefal to do that you need to hook in `pause` cordova event. We use already the `resume` handler in OnlineRound.ts so you have an example.

freefal:
Ok, thanks for the tip.

freefal:
Hmmm...the asynchronous nature of webworkers doesn't easily facilitate our desire to guarantee that the 'bye' message gets sent before the 'pause' event finishes. Need to think about how to get 'pause' to wait for the 'bye' message to get sent or implement another way.

veloce:
Done it here: https://github.com/veloce/lichobile/commit/126cc3ae50101caaa1de59f6a5ea265b8c0ea541

Happy0:
yayyy

Issue #412, opened 2016-09-01
PGN piece notation preference


ornicar:
would be a nice addition.

See https://lichess.org/account/preferences/game-display

PR #446, opened 2016-10-11
Add option to show piece notation to move list in analysis view

freefal:
Closes https://github.com/veloce/lichobile/issues/412


veloce:
Thanks for that!

It would be even nicer if this was a global settings, under 'settings -> game display' entry probably, and would display the piece notation also in game replay (visible in landscape mode)

freefal:
It is in 'settings -> game display'. I'll make game replay use it as well.
I never go into landscape so didn't know it was used elsewhere.

On Tue, Oct 11, 2016 at 3:38 AM, Vincent Velociter <notifications@github.com

veloce:
One more location to change: offline games in landscape mode.

project/src/js/ui/shared/round/view/replay.js
```diff
@@ -1,4 +1,5 @@
 import * as helper from '../../../helper';
+import settings from '../../../settings';
```
veloce: bad path

veloce:
You could cache access to the pref since it's been read a lot during redraws and it's localstorage.

See how I do it in `Board.tsx` for pieceTheme and boardTheme as an example.

freefal:
Apologies for the bad settings. I thought I tested everywhere I added the
pieceNotation, but clearly I didn't. Now fixed.
Thanks for the tip on caching. Now implemented.

On Sat, Oct 15, 2016 at 1:11 PM, Vincent Velociter <notifications@github.com

PR #450, opened 2016-10-15
Bug fix: when searching for players, the list did not redraw when new…

Happy0:
… results were found. Fixed by adding a 'redraw' each time the results are refreshed after the XHR request to autocomplete.


Happy0:
Also, my change makes the client only ask the server for the autocomplete list after 3 characters have been typed as the server only starts responding after 3 characters:

https://github.com/ornicar/lila/blob/master/modules/user/src/main/UserRepo.scala#L256

veloce:
Thanks!

Happy0:
yayyy

PR #460, opened 2016-10-17
Bug fix: scroll chat to bottom when it is opened and when it is updat…

Happy0:
…ed on sending or receiving a message.

### DIFF
--- a/project/src/js/ui/shared/round/chat.js
+++ b/project/src/js/ui/shared/round/chat.js
@@ -56,7 +56,7 @@ export function chatCtrl(root, isShadowban) {
   }.bind(this);
 
   this.scrollChatToBottom = function(el) {
-    el.scrollTop =  el.scrollHeight;
+    el.scrollTop = el.scrollHeight;
   }.bind(this);
 
   function onKeyboardShow(e) {
--- a/project/src/js/ui/shared/round/chat.js
+++ b/project/src/js/ui/shared/round/chat.js
@@ -55,6 +55,10 @@ export function chatCtrl(root, isShadowban) {
     redraw();
   }.bind(this);
 
+  this.scrollChatToBottom = function(el) {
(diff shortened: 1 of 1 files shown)


Happy0:
I tried it with the keyboard enabled in google chrome developer tools, but I don't know if that's an accurate representation of how it will behave in the app.

I guess I'll try to get it on my phone tomorrow / soon just to make sure.

project/src/js/ui/shared/round/chat.js
```diff
@@ -55,6 +55,10 @@ export function chatCtrl(root, isShadowban) {
     redraw();
   }.bind(this);
 
+  this.scrollChatToBottom = function(el) {
+    el.scrollTop = 999999;
```
veloce: `el.scrollTop = el.scrollHeight` is supposed to work, why did you change it?

Happy0:
It doesn't work for me on android. Maybe you need el.dom.scrollHeight = ?.

el is not a Dom element but has a 'don't property

veloce:
Ah... yes you're right, it's a regression. `el` here is the vnode, not the dom element.
So yes it should be: 

```js
this.scrollChatToBottom = function(el) {
  el.scrollTop = el.scrollHeight;
}
...

oncreate: (vnode) => this.scrollChatToBottom(vnode.dom)
onupdate: (vnode) => this.scrollChatToBottom(vnode.dom)
```

Happy0:
Will do! 'el.scrollHeight' is better than = 9999999 :P.

Happy0:
I made the changes using github's web interface, will test when i get home later.

Happy0:
I've tested this both in chrome dev tools and as an android APK (including with the keyboard open.) Seems to work fine!

Feel free to merge if you're happy with it.

Issue #474, opened 2016-10-28
Advanced game search


veloce:

sepiropht:
Can i work for this one ? What should i know to start

Issue #394, opened 2016-08-05
Offline variants help


veloce:

veloce:
Since we have all variant playable offline it would be nice to have the same help pages [the website has](https://fr.lichess.org/variant/crazyhouse) but embedded in the app to be seen offline.

Help would be displayed in a modal window (like the one for the tournament FAQ).

gurditsbedi:
i would like to contribute to this feature. I have good experience with javascript(es6, react, react-native). I have no experience with Cordova. How much difficult will it be to contribute?

veloce:
Great! If you have good experience with javascript and react there is no difficulty to write this feature in my opinion. You'll want to read more about [mithril](https://github.com/MithrilJS/mithril.js) which has some differences with react but is also a VDOM lib with components. Cordova is not an issue here.

Quick summary of how I see this feature:
- [ ] use [tournament faq](https://github.com/veloce/lichobile/blob/master/src/ui/tournament/faq.tsx) as a basis to create a mithril component that shows a static html help page (take the layout and sytle of it)
- [ ] this component provides the modal, content is passed with the `chilldren` prop
- [ ] take the html from [the website](https://lichess.org/variant) (don't include videos)
- [ ] add a "Rules" button in both AI and OTB screens menu, which shows the modal

gurditsbedi:
So,
Should the information of all chess variants be in rendered in a single modal?


Where should I keep this file? It is going to be referenced by `ai` and `otb` screen.
I am thinking of `src/ui/chessVariantsFaq.tsx`

Where should be the trigger button? I think this information should be in offline and also online modes?
Going by your idea... The trigger button should be in the start screen(which comes when app is in offline mode), below the buttons `computer` and `over the board`.?

veloce:
There should be one HTML (mithril template) file per variant I think. Then the help modal composant requires the proper file according to the `variant` prop. So there is one help page per variant.

You're right, this help should also comes with online games. I think the game menu is a good place: 
![screenshot from 2018-01-08 15-50-38](https://user-images.githubusercontent.com/423393/34676060-ccb8a8c2-f48b-11e7-9664-1decf394e55b.png)

gurditsbedi:
As per [line 10](https://github.com/veloce/lichobile/blob/master/src/ui/tournament/faq.tsx#L10) there is this `interface` parameter `root`. What will be the equivalent for the helper variant modal?

veloce:
[...]

PR #482, opened 2016-11-06
add endgames positions

sepiropht:


www/data/endgames.json
```diff
@@ -0,0 +1,657 @@
...
+    "fen": "7k/R7/7K/8/8/1b6/8/8 w - - 0 1",
+    "name": "Basic ending XVIII"
+}, {
+    "eco": "18",
+    "fen": "6k1/5R2/6K1/8/8/8/8/6b1 w - - 0 1",
+    "name": "Basic ending XX"
+}, {
+    "eco": "19",
+    "fen": "8/8/8/8/8/3k4/r7/3NK3 w - - 0 1",
+    "name": "Basic ending XX1"
```
veloce: small typo here

src/js/ui/editor/menu.js
```diff
@@ -80,6 +80,24 @@ export function renderSelectColorPosition(ctrl) {
         )
       ])
     ]),
+    m('div.select_input', [
+      m('label', {
+        'for': 'select_editor_positions'
+      }, 'Endgames'),
+      m('select.positions', {
+        id: 'select_editor_positions',
```
veloce: The `id` must be different that the other selector's one.

src/js/ui/editor/menu.js
```diff
@@ -80,6 +80,24 @@ export function renderSelectColorPosition(ctrl) {
...
+        onchange: function(e) {
+          ctrl.loadNewFen(e.target.value);
+        }
+      }, [
+        optgroup('Set the board', [
```
veloce: You should add the empty `-- Position --` item like in the opening selector

sepiropht: All changes done ! 
I have a question, how do you launch the project in browser ? Last week is worked well, 
but now after 'npm run watch', i have to do '  python -m SimpleHTTPServer 8080' 
If not, anything happen in browser. 
Even after that, it work, but i still have s cors issue and i'm not able to make any request

sepiropht: I said 'nothing happens' but I should say that I get this error in the console
'Uncaught SecurityError: Failed to construct 'Worker': Script at 'file:///home/sepiropht/src/lichobile/www/lib/socketWorker.js' cannot be accessed from origin 'null'.'

sepiropht: Ok it working like before. I think that line ' chromium-browser --user-data-dir=$HOME/.chromium_dev_dir --disable-web-security
 ' should remains in the readme shouldn't it ?

veloce: yes that's the way to disable security in chrome

veloce:
Thanks!

sepiropht:
Yes. I would like to do more. What else can I do? I'm looking for a feature or a bug, but nothing seems obvious

veloce:
Contributions are appreciated! Current bugs are not obvious, indeed. Instead I can suggest you one of the following features: #474, #473, or maybe #394 (should be the easiest).
Let me know if one of those suits you, and if you need help. Note that I'm currently in the process of migrating code to typescript, so any consequent new PR should be written in typescript.

sepiropht:
Ok thanks for the response. I will look at them

sepiropht:
I will try the first feature(game search), if it is too difficult, I will take the last

Issue #236, opened 2015-12-16
Local notification when time is low in correspondance games


veloce:

JoelMon:
I vote for this enhancement. Also would like if the user can set when to receive the notification (within 24hrs, 30min, etc) before the time ends.

PR #485, opened 2016-11-08
Handle 'corresAlarm' push notifications for correspondence players who are low on time

freefal:
Closes https://github.com/veloce/lichobile/issues/236

### DIFF
--- a/src/js/push.js
+++ b/src/js/push.js
@@ -16,13 +16,15 @@ function notificationReceivedCallback(data) {
           window.plugins.toast.show(
             i18n('userAcceptsYourChallenge', lightPlayerName(additionalData.userData.joiner)), 'long', 'top');
           break;
-        case 'newMessage':
-          window.plugins.toast.show(
-            'New message from ' + lightPlayerName(additionalData.userData.sender), 'long', 'top');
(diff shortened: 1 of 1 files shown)


ornicar:
are you done testing this on stage? Can I stop sending the message 10s after the move?

freefal:
No, sorry I should have marked this "[To be tested]". Please leave stage as it is. I used to know how to enable push on dev builds when we were using tarifa, but don't know how to do it with vanilla cordova. I've pinged @veloce in #mobile to ask him how to do it now.

freefal:
@ornicar Ok, this is now confirmed to work. When the app is not in the foreground, the user gets the push notification and it links to the game as expected.

Weirdly, I am not getting any push notification when the app is the foreground. I would have liked to display a toast here (https://github.com/freefal/lichobile/blob/6488e21b652f0a25c7a63816d173400ba7fce304/src/js/push.js#L19), but the code never reaches there when the app is the foreground. I tried adding some `console.log`s above all the ifs and saw nothing for moves confirmations or low time warnings. 

I don't understand, but not a big deal. The core functionality is there.

PR #486, opened 2016-11-08
Fix tournament move warning when spectator

freefal:
Closes https://github.com/veloce/lichobile/issues/452

### DIFF
--- a/src/js/ui/shared/round/view/roundView.jsx
+++ b/src/js/ui/shared/round/view/roundView.jsx
@@ -1,0 +1,382 @@
+"use strict";
+var m = require("mithril");
+var chessground = require("chessground-mobile");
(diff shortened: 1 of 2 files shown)


src/js/ui/shared/round/view/roundView.jsx
```diff
@@ -0,0 +1,382 @@
+"use strict";
```
veloce: oops file added by mistake

freefal: Sorry about that. Ever since these files started showing up I've been adding specific files or doing a `git diff` before a `git add -A` to make sure there are none. Not sure how this one slipped through. Will be more vigilant going forward.

PR #491, opened 2016-11-15
Update tablebase endpoint

niklasf:
There's a new tablebase endpoint.

Changes:

* Moved to new server that has space for variant tablebases in the future
* Start querying the 6 man tablebases when there are 7 pieces on the board to display winning transitions
* API sends less unused information, `real_wdl` -> `wdl`

We can keep the old endpoint up as long as we like.


veloce:
Thanks!

PR #498, opened 2016-11-22
Deep linking without custom url scheme

freefal:


config.xml
```diff
@@ -51,9 +51,12 @@
     <preference name="StatusBarOverlaysWebView" value="false" />
     <preference name="StatusBarBackgroundColor" value="#151A1E" />
     <preference name="BackgroundColor" value="0xFF151A1E" />
-    <preference name="xwalkVersion" value="22" />
+    <preference name="xwalkVersion" value="20" />
```
veloce: I guess it has been automatically added. Please remove all crosswalk related changes here since I already made changes

ul_web_hooks/android/android_web_hook.html
```diff
@@ -0,0 +1,25 @@
+<!--
```
veloce: Why is that file useful?

freefal:
It was auto-generated by the new cordova plugin. It's a bit annoying that
it gets generated in the top-level folder.

It certainly doesn't belong in the source tree. I should have added it to
`.gitignore` and will do so now.

On Tue, Nov 22, 2016 at 5:01 AM, Vincent Velociter <notifications@github.com

freefal:
I will revert to version 22. I couldn't get it to compile last night for
some reason (it was timing out on some maven repository). Hopefully it will
work for me today.

On Tue, Nov 22, 2016 at 4:59 AM, Vincent Velociter <notifications@github.com

src/js/deepLinks.ts
```diff
@@ -29,9 +30,63 @@ for (let dr in deepRoutes) deepRouter.add(dr, deepRoutes[dr]);
...
+  const pieces = eventData.path.split('/');
+  const uid = pieces[2];
+  const variant = pieces[4];
+  router.set('/@/' + uid + '/' + variant + '/perf');
+}
+
+function handleTrainingProblem (eventData: EventData) {
+  const pieces = eventData.path.split('/');
+  const problem = pieces[2];
+  if (problem === 'coordinate') {
```
veloce: Can't we just avoid to open coordinate training in an external window? it doesn't make sense, we should just ignore it and leave it in the website.

src/js/deepLinks.ts
```diff
@@ -29,9 +30,63 @@ for (let dr in deepRoutes) deepRouter.add(dr, deepRoutes[dr]);
 
 export default {
   init() {
-    // open games from external links with url scheme
-    window.handleOpenURL = function(url: string) {
+    // open links that follow the custom url scheme
+      window.handleOpenURL = function(url: string) {
```
veloce: Why keeping both `deepRouter` and `universalLinks` routers? If I understand correctly, universalLinks is a replacement of the older solution. It seems it will create conflicts if both stays.

freefal:
Here's a nice little test suite of links

Analysis: https://en.lichess.org/analysis (app)
Challenge: Create your own (app)
Editor: https://en.lichess.org/editor (app)
Inbox: https://en.lichess.org/inbox (app)
New Message: https://en.lichess.org/inbox/new (app)
Players: https://en.lichess.org/player (app)
Game: https://en.lichess.org/b9G5UVbM (app)
Tournament: https://en.lichess.org/tournament/QsEkR4FO (app)
Tournament List: https://en.lichess.org/tournament (app)
Training Problem: https://en.lichess.org/training/4760 (app)
Training: https://en.lichess.org/training (app)
TV: https://en.lichess.org/tv (app)
TV Channel: https://en.lichess.org/tv/bullet (app)
Player: https://en.lichess.org/@/RebeccaHarris (app)
Player Perf: https://en.lichess.org/@/RebeccaHarris/perf/bullet (app)

As a first step, nothing needs to change on the server. We are simply
enabling better handling of normal lichess links by the app. Once we've
deployed this and users have widely upgraded, then we can explore replacing
the server-side links with these links. I don't think there is any conflict
because the `universalLinks` library is handling the `https://` scheme and
`rlite-route` is handling the custom `lichess://` scheme.

On Sun, Nov 27, 2016 at 9:51 AM, Vincent Velociter <notifications@github.com

src/js/deepLinks.ts
freefal: The issue is that the `universalLinks` cordova plugin does not have regex support. We can only match paths against a simple `*` wildcard. Because games, which we definitely want to give the user the ability to direct to the app, are of the form `https://en.lichess.org/[8 letters or numbers]`, we have to handle the case of `https://*.lichess.org/*`, test to see if it's a game with javascript regex matching code, and then direct back out to a browser if not.
The same applies the training. We have to match `https://*.lichess.org/training/*` and then direct the user back out to the web browser if `*` turns out to be `coordinate` rather than some problem number.

veloce:
Ok I need to test it on iOS. I trust you for android. 
If I understand things correctly, those links will be useful only when their references are added in the website, right?

The test suite is important, can you add it to the existing `deepLinkTest.html` file?

freefal:
No, they will be immediately be useful. Right now if I found a cool problem
(e.g., `https://en.lichess.org/training/61062`) and sent it to you and you
opened it in your phone's email app, that link would open in your mobile
browser. With my pull request, you will have the option to open that link
directly in the app (on Android, you are given the option between app and
mobile browser the first time and your preference is saved thereafter).

However the `lichess://` scheme is still used by those big blue and
white buttons that are presented at the top of the mobile browser page when
you navigate to a game in a browser. Once this pull request is merged and
released, we can consider changing those links to be normal `https://`
links with some hint that causes them to open in the app. Then we can
discard the cordova plugin that enables the `lichess://` scheme.

I will add the test suite now.

On Sun, Nov 27, 2016 at 3:28 PM, Vincent Velociter <notifications@github.com

config.xml
```diff
@@ -54,6 +54,9 @@
     <preference name="xwalkVersion" value="22" />
     <preference name="xwalkCommandLine" value="--enable-viewport-meta --disable-pull-to-refresh-effect" />
     <preference name="android-minSdkVersion" value="16" />
+    <preference name="xwalkMode" value="embedded" />
```
veloce: Those 3 added preference should be removed from the diff. They are implicit.

veloce:
@freefal I've tested this on my android emulator with the test suite in html file, and it does nothing when I click a link (ie. it redirects to the website as expected).

freefal:
Ok, I will remove.

On Mon, Dec 5, 2016 at 12:45 PM, Vincent Velociter <notifications@github.com

freefal:
Hmmm....is that true even when you do a fresh clone with a fresh `npm
install` and `cordova platform add android`? It works fine for me on my
phone.
I'm wondering if somehow the new cordova plugin isn't ending up in your
build.



On Mon, Dec 5, 2016 at 12:46 PM, Vincent Velociter <notifications@github.com

veloce:
Normally the plugin is downloaded during build if not present, and I saw the plugin install logs. But, yeah, I will retry with a fresh clone.

freefal:
Huh, if you saw it in the plugin install logs then it probably is there.
Not sure. Please let me know if a fresh clone fails and I'll see if I can
reproduce on my physical phone and then the android emulator.

veloce:
@freefal I've tested again. It works, there's one caveat though: the link will open the right screen only if the app is closed. When the app is already open, clicking the link just put it in the foreground without routing internally to the corresponding screen.

PR #501, opened 2016-11-27
Rating Graph for Variant Pages

freefal:
Also refactor usage of chart.js withing training module to work with chart.js 2.4

Closes https://github.com/veloce/lichobile/issues/473


src/js/ui/user/variantperf/variantPerfView.js
```diff
@@ -169,7 +173,104 @@ function isEmpty(element) {
...
     return str.replace(/\w\S*/g, function(txt){return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();});
 }
+
+function createGraph(data, node) {
+  const ctx = node.dom.getContext('2d');
+  drawChart(data, ctx);
+  node.state.hash = chartHash(data);
+}
+
+function updateGraph(data, node) {
```
veloce: I think the whole `updateGraph` logic is useless here, since the perf screen don't auto update. Each time you go again to the screen it will be updated, so only the `oninit` logic is needed.

src/js/ui/user/variantperf/variantPerfView.js
```diff
@@ -14,140 +17,141 @@ export default function view(vnode) {
...
+
+  if (!data) return null;
+
+  const days = Math.floor(data.stat.count.seconds / (60 * 60 * 24));
+  const hours = Math.floor(data.stat.count.seconds / (60 * 60)) - days * 24;
+  const mins = Math.floor(data.stat.count.seconds / 60) - days * 24 * 60 - hours * 60;
+
+  return (
+    <div class="variantPerfBody native_scroller page">
+      <canvas oncreate={node => createGraph(data, node)} onupdate={node => updateGraph(data, node)} />
```
veloce: I think you can just add a `{ key: isPortrait ? 'portrait' : 'landscape' }` to force canvas recreation on orientation change, and avoid the whole `onupdate` logic.

freefal: Agreed. I will fix.

PR #507, opened 2016-12-05
Endgame positions training

sepiropht:
I have an issue with this , the list doesn't show up on start


sepiropht:
no issue anymore,

src/js/routes.ts
```diff
@@ -44,6 +45,7 @@ export default {
       'otb': otb,
       'otb/fen/:fen': otb,
       'ai': ai,
+      'eac': eac,
```
veloce: what does eac mean?

src/js/ui/eac/index.jsx
```diff
@@ -0,0 +1,44 @@
+"use strict";
```
veloce: the compiled `.jsx` file should not be there.

src/js/ui/eac/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </ul>
+      );
+    }
+
+    function start() {
+      settings.gameSetup.ai.level('8');
+      settings.gameSetup.ai.variant('3')
+      settings.gameSetup.ai.color('white');
+      function startAIGame() {
+        return xhr.newAiGame(ctrl.position)
```
veloce: This should stay an offline training mode, against the stockfish embedded in the app.

src/js/ui/eac/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    function renderBody() {
+      return (
+        <ul className="native_scroller page settings_list radio_list">
+          {ctrl.endgamesPositions().map((p: Pos) => renderPosition(p))}
+        </ul>
+      );
+    }
+
+    function start() {
```
veloce: such a function should not be defined in `view` function of component, rather in `oninit`

src/js/routes.ts
sepiropht: endgame against computer

src/js/ui/eac/index.jsx
sepiropht: ok i will remove it

src/js/ui/eac/index.tsx
sepiropht: ok i will move it in te on iniit

src/js/ui/eac/index.tsx
sepiropht: there is an option missing ? this one settings.gameSetup.ai.availableVariants[1] ?

src/js/ui/eac/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+
+      function start() {
+        settings.gameSetup.ai.level('8');
+        settings.gameSetup.ai.variant('3');
+        settings.gameSetup.ai.color('white');
+        settings.gameSetup.ai.availableVariants[1];
+        
+        function startAIGame(position:string) {
+          return xhr.newAiGame(position)
```
veloce: It has to be an offline game. So instead of making an xhr, it should use instead the `AiRound` module.

Either it redirects to ai screen (`router.set('/ai/fen/fenstring')`) or better it loads directly the `AiRound` class (which needs to be refactored for that purpose, and not be coupled to global settings anymore).

sepiropht: ok i will do that

PR #526, opened 2016-12-28
Allow import as FEN.

therealmitchconnors:
Many Chess OCR apps (i.e. the /r/chess fenbot) can produce FEN notation from a chess diagram, allowing users to quickly import tactical puzzles and chess book examples.  To support this in the mobile app, we intercept a failed PGN import, wrap the notation with a FEN tag in the PGN format, and retry.  As noted below, it would be even better if the server-side API could identify FEN vs PGN vs EPD, etc.


therealmitchconnors:
Also, would there be any interest in adding an chess computer vision feature that would allow directly importing a chess diagram?  If I adapted the fenbot to javascript, would you consider merging that feature in?  It would mean that the app would need to request camera permissions...

therealmitchconnors:
I have improved this commit to parse and resolve FEN strings without any "try and retry" logic against the API, which slows the app, and negatively impacts the API server.  FENs are handled client-side now, and I have updated the i18n string in English to reflect this.  What is the process for getting the other languages changed?

therealmitchconnors:
In my most recent commit, I have added support for deeplinks such as https://en.lichess.org/analysis/8/8/1r3K1R/5P2/3p3p/4p1P1/3k4/8_w_- , which is a common format on reddit and other forums.

src/js/ui/importer/index.ts
```diff
@@ -106,4 +112,16 @@ function renderBody(ctrl: State) {
   ]);
 }
 
+function fetchImport(data: SendData) {
```
veloce: this function is never used

www/i18n/en.json
```diff
@@ -1 +1 @@
...ardToMoveInTheGame":"You can also scroll over the board to move in the game.","pressShiftPlusClickOrRightClickToDrawCirclesAndArrowsOnTheBoard":"Press shift+click or right-click to draw circles and arrows on the board.","confirmResignation":"Confirm resignation","letOtherPlayersMessageYou":"Let other players message you","shareYourInsightsData":"Share your insights data","youHaveAlreadyRegisteredTheEmail":"You have already registered the email: %s","kidMode":"Kid mode","playChessEverywhere":"Play chess everywhere","asFreeAsLichess":"As free as lichess","builtForTheLoveOfChessNotMoney":"Built for the love of chess, not money","everybodyGetsAllFeaturesForFree":"Everybody gets all features for free","zeroAdvertisement":"Zero advertisement","fullFeatured":"Full featured","phoneAndTablet":"Phone and tablet","bulletBlitzClassical":"Bullet, blitz, classical","correspondenceChess":"Correspondence chess","onlineAndOfflinePlay":"Online and offline play","correspondenceAndUnlimited":"Correspondence and unlimited","viewTheSolution":"View the solution","followAndChallengeFriends":"Follow and challenge friends","availableInNbLanguages":"Available in %s languages!","gameAnalysis":"Game analysis"}
```
veloce: Don't edit that file, it is fetched from the server and overwritten

veloce:
Sorry for the late comment. This is interesting as it adds some features like opening analysis position from the website. 

But regarding the fen import, I don't understand exactly the problem you're trying to solve and why one would paste a FEN string in the import textarea which states clearly "paste PGN here".

If the problem is that some generated pgn is not recognized by lichess server, I suggest to open an issue in ornicar/lila repo. Maybe you can describe more precisely there what part of that PGN is not well handled by the server. If there is a way to improve importer server-side it would benefit everyone.

veloce:
Actually now I that I read again your first comment, I see that you really want to quickly load a fen position in the app. I did not understand that in the first place sorry. 
I created an issue for that: https://github.com/veloce/lichobile/issues/541

If you want to do it, you can modify your PR and leave only the deep link part and the fen loading stuff in the editor screen, leaving the import screen untouched. What do you think?

src/js/ui/importer/index.ts
therealmitchconnors: Sorry, that was a direction I abandoned.  I will remove it.

therealmitchconnors:
Thanks for the review!  I think in hindsight that the import screen was the wrong place to put this feature.

There are four FEN-related use cases that I think would make the app much more useful, especially when studying a chess book.  Each of them basically takes a FEN as input, and loads the Analysis page with that position.

1. Copy/Paste FEN (accomplished here with "Import", perhaps belongs somewhere else?)
2. Click lichess.org/analysis/fen link (accomplished here)
3. "Share" a FEN from another Chess-related app (I'm trying to get this to work with cordova-intents, but something is not working correctly).
4. Take a picture of a chess board or diagram, and generate a FEN using tensorflow (I am working on porting ChessFenBot to Java for this purpose)

Is there a better place for the copy/paste FEN logic?  If I am able to accomplish the "take a picture" logic, where would you like to see that in the UI?

veloce:
About item 1. of your list, I think the best place is in the board editor screen: just add a button on the bottom tab bar. It should display a popup with a text input, you paste a fen and it loads the position.

If you want I can do it. You should then remove from this pr everything related to the import and keep only the deeplinking part. So I could merge it.

About 3., I've also never done that, but I would accept a PR of course! (please keep PRs separated by feature)

About 4., same thing: you can send me a PR :) This "take a picture" logic belong to the board editor as well actually. It could be another button on the bottom bar (camera icon).

Actually if you're going to implement such a feature in the board editor. I'd be nice to have it rewritten in typescript. I'll try to do that asap. Please stay in sync so you can write this feature in typescript. Thanks!

therealmitchconnors:
OK, I just rolled back those importer changes.  Somewhere along the lines, something went wrong with my git log, so there are a few superfluous commits.  Will those get squashed in the merge?  

I am on vacation this week, so my responses will be delayed.  I think putting the paste feature into a type-script board editor sounds awesome.   I will continue my work on porting the chess OCR tensorflow model to Java, so we can import it as a cordova plugin.  Once I have something worth showing, I'll send you a link to the repo...

PR #536, opened 2017-01-19
Enable antichess tablebases

niklasf:
We also have antichess tablebases now.

### DIFF
--- a/src/js/ui/analyse/interfaces.ts
+++ b/src/js/ui/analyse/interfaces.ts
@@ -227,6 +227,8 @@ export interface ExplorerMove {
   checkmate: boolean
   stalemate: boolean
   insufficient_material: boolean
+  variant_win: boolean
+  variant_loss: boolean
 }
 
 export interface ExplorerPlayer {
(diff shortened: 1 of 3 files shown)


src/js/ui/analyse/interfaces.ts
```diff
@@ -251,4 +253,6 @@ export interface ExplorerData {
   recentGames?: Array<ExplorerGame>
   checkmate?: boolean
   stalemate?: boolean
+  variant_win?: boolean
+  variant_loss?: boolean
```
ornicar: why the coding standard change? can we stick to `camelCase` here?

niklasf: This describes the tablebase JSON API. Unfortunately it already uses `insufficient_material`. But at least it's self consistent.

ornicar:
oopsies, I thought I was merging on lila.

veloce:
lol, looking at the code I think there isn't too much risk. I'm gonna test right now though

veloce:
confirm it's working

PR #560, opened 2017-02-20
Advanced search

freefal:
Closes https://github.com/veloce/lichobile/issues/474


src/js/ui/search/searchCtrl.ts
```diff
@@ -0,0 +1,82 @@
...
+    .then((data: SearchResult) => {
+      result(prepareData(data));
+      games(games().concat(result().paginator.currentPageResults));
+      redraw();
+    })
+    .catch(handleXhrError);
+  }
+}
+
+function buildQuery (elements: HTMLCollection, acc: any, name: string) {
```
veloce: I think the `acc` type should be `Partial<SearchQuery>` ;
In almost all cases, there is no need to use `any`.

src/js/ui/search/searchView.tsx
```diff
@@ -0,0 +1,238 @@
...
+      onsubmit={function(e: Event) {
+        e.preventDefault();
+        const analysed = document.getElementById('analysed') as HTMLInputElement;
+        analysed.value = analysed.checked ? '1' : '';
+        return ctrl.search(e.target as HTMLFormElement);
+      }}>
+          <div className="game_search_row">
+            <label>Players: </label>
+            <div className="game_search_input">
+              <input type="text" id="players_a" name="players.a" onkeyup={(e: Event) => { redraw(); }} />
```
veloce: prefer `oninput` to `onkeyup`. No use to declare function arg if you don't use it: `() => redraw()` is simpler

it is also useful to throttle this input handler

src/js/ui/search/searchView.tsx
```diff
@@ -0,0 +1,238 @@
...
+function playersNonEmpty() {
+  return getPlayers().reduce((acc, cur) => acc + cur.label.length, 0) > 0;
+}
+
+function isComputerOpp() {
+  const hasAi = (document.getElementById('hasAi') as HTMLInputElement);
+  return hasAi && (hasAi.value === '1');
+}
+
+const Game: Mithril.Component<{ g: UserGameWithDate, index: number, boardBounds: {height: number, width: number}, ctrl: SearchState }, { boardTheme: string }> = {
```
veloce: This is the same component as in userGames.tsx. It should be extracted, put in `shared` folder.

veloce:
Thanks! this is good, but it would be better if the search was remembered as in the website. When you search sth, update the url to have a permalink to the search, so when you go to a game and go back, you go back to your search.

src/js/ui/search/searchView.tsx
freefal: The components aren't quite the same. The concept of `scrollState` does not exist for the `Game` object used in the search view.

Issue #562, opened 2017-02-24
Importing a concluded OTB game to lichess.org


arex1337:
After having played a game in the OTB mode in the app, and assuming I'm logged in and online (anon support and offline support could be considered, but not a big priority for me), I would like to easily import/save the game to lichess.org. I can then go to lichess.org to analyze the game later.

Since White and Black does not have usernames in OTB mode, the PGN that is imported into lichess.org should contain (real world) clock time. This will make it easier to identify the game, assuming many games are played and imported in a row.

freefal:
@arex1337 I couldn't easily add the time as our PGN dump currently uses scalachessjs, which only includes a date. If @veloce adds a [Time HH:MM:SS] field to the PGNDump function in scalachessjs (https://github.com/veloce/scalachessjs/blob/master/src/main/scala/scalachessjs/PgnDump.scala)  which Wikipedia says is a common tag (https://en.wikipedia.org/wiki/Portable_Game_Notation), then we'll have the time in the PGN that the app creates.
For the time being, you can manually add it if you wish.

veloce:
I think we'll do better than that. I'll include the date and time (I'll modify scalachessjs). But when importing the PGN to lichess, I was thinking that we could prompt user to write some metadatas that would go along the upload. Not too much and all optional of course. Like:

- black player
- white player
- site

What do you think @arex1337 ?

PR #567, opened 2017-03-02
Add ability to import over-the-board games to Lichess

freefal:
Closes https://github.com/veloce/lichobile/issues/562

### DIFF
--- a/src/js/settings.ts
+++ b/src/js/settings.ts
@@ -152,7 +152,7 @@ export default {
     useSymmetric: localstorageprop('settings.otb.useSymmetric', false),
     variant: localstorageprop('settings.otb.variant', 'standard'),
     availableVariants: offlineAvailableVariants,
-    savedGames: localstorageprop('settings.otb.savedGames', {})
+    savedPGN: localstorageprop('settings.otb.savedPGN', '')
   },
 
   clock: {
--- a/src/js/settings.ts
+++ b/src/js/settings.ts
@@ -151,7 +151,8 @@ export default {
     flipPieces: localstorageprop('settings.otb.flipPieces', false),
     useSymmetric: localstorageprop('settings.otb.useSymmetric', false),
     variant: localstorageprop('settings.otb.variant', 'standard'),
-    availableVariants: offlineAvailableVariants
+    availableVariants: offlineAvailableVariants,
+    savedGames: localstorageprop('settings.otb.savedGames', {})
   },
 
   clock: {
--- a/src/js/ui/otb/actions.js
+++ b/src/js/ui/otb/actions.js
@@ -53,10 +53,8 @@ export default {
       },
       importGame: function() {
         root.replay.pgn().then(data => {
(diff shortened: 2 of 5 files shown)


src/js/settings.ts
```diff
@@ -151,7 +151,8 @@ export default {
     flipPieces: localstorageprop('settings.otb.flipPieces', false),
     useSymmetric: localstorageprop('settings.otb.useSymmetric', false),
     variant: localstorageprop('settings.otb.variant', 'standard'),
-    availableVariants: offlineAvailableVariants
+    availableVariants: offlineAvailableVariants,
+    savedGames: localstorageprop('settings.otb.savedGames', {})
```
veloce: I don't want to store all played games right now. And not with localstorage. Localstorage space is limited and if we do that, we'd want to add the ability to delete past games.

And I don't get why you do this, since you always use the same id, if I get it correctly (`1`)

src/js/ui/otb/actions.js
```diff
@@ -7,6 +7,7 @@ import ground from '../shared/offlineRound/ground';
 import popupWidget from '../shared/popup';
```
veloce: Please rewrite this file in typescript if you modify it.

src/js/settings.ts
freefal: @veloce My idea here was to create a general data structure that we could use in the future if we decided to save all games (at which point the object keys would be gameIds), but to just hardcode it to 1 for now for the exactly the reason you described.

I'm happy to just turn this into a string holding the single pgn though. We can change later if we start saving games.

src/js/ui/otb/actions.js
freefal: Haha, conscripted into the typescript army! Ok, I will rewrite in typescript.

veloce: haha :) 

When I'm done with chessground (already rewritten by @ornicar ) and puzzles we'll have almost all lichobile in typescript. That will be much better, since as long you keep some javascript you always risk to break something

freefal:
While I still sometimes get annoyed when I need to stop and describe my
data structures via interfaces, I now spend more time fixing bugs at
compile time and less at runtime, which is definitely a good thing.

On Thu, Mar 2, 2017 at 10:02 AM, Vincent Velociter <notifications@github.com

freefal:
Ok, I think both of your comments on this PR have been incorporated.

src/js/ui/otb/actions.js
markyi370: What's the difference between Typescript/Javascript?

veloce:
Sorry I'm closing it since it has diverged from master and I want to do this feature differently (there will be a prompt to enter user names before submitting to lichess)

freefal:
Understood. Happy to see the feature get incorporated in whatever form
makes sense.

On Tue, May 16, 2017 at 3:10 PM, Vincent Velociter <notifications@github.com

PR #568, opened 2017-03-03
Return only friends from search until no friends match. Then search everyone.

freefal:
The `friend` flag is a neat option that @ornicar showed me. It's a nice way to prioritize what most users are doing most of the time (searching for a friend), but still pulling up other results when the user enters a prefix that doesn't match any of their friends.


veloce:
Nice, indeed.

PR #608, opened 2017-06-07
Implem new move time

veloce:


ornicar:
This PR could be renamed. This is not the timeseal protocol.

`Implement new move time` would work.

veloce:
ok. I thought it was similar to timeseal, even if not a protocol

PR #612, opened 2017-06-10
training: include current rating in graph

### LINKED ISSUE #611: training rating graph is missing current rating
The puzzle rating history shows the current rating in the graph title, but not as part of the graph series.

christophgysin:
fixes #611


veloce:
Thanks for the catch!

PR #625, opened 2017-07-18
Rewrite analysis using lichess Tree module

veloce:


src/js/ui/analyse/AnalyseCtrl.ts
```diff
@@ -207,7 +207,7 @@ export default class AnalyseCtrl {
       orientation: this.vm.flip ? oppositeColor(this.orientation) : this.orientation,
       movableColor: this.gameOver() ? null : color,
       dests: dests || {},
-      check: node.check,
+      check: node.check !== undefined ? node.check : false,
```
ornicar: is it equivalent to `check: !!node.check`?

veloce: yes :) I should change that

PR #626, opened 2017-07-25
Update lag protocol for app

isaacl:
- Send lag in pongs with same logic as website
- Don't bother with lag millis if sending move time.
- Use improved formula for calculating average lag.


veloce:
@isaacl I'm now available to test it. In 3 weeks maybe the code has changed even more on the site. Is there more to do related to lag protocol?

isaacl:
This specific PR is pretty stable, but there's a couple related features in flux:
- opponent lag prediction: opp's clock pauses before starting after your move, based on a server value
- showing lag bar charts for opp

veloce:
Thanks!

PR #650, opened 2017-09-09
Clock improvements

### LINKED ISSUE #647: clock enhancements for app
- ensure flagged message goes out when clock reaches 0
- socket 'withLag' should be true for first two moves
- delay opponent clock based on lag est from server

isaacl:
- Reset move timer on reload
- Send lag values while clocks are paused
- Use lag value for delay

This fixes #647.


veloce:
Thanks!

Issue #99, opened 2015-02-12
Support timed offline games


tobiasvl:
It would be great if offline games had the option to have a clock, so you don't have to have a separate clock (or even a separate smartphone with a chess clock app). It should probably have a pause button.

garawaa:
It s good idea. Then upper side players clock should be flipped in over the board game.

wei2912:
Any updates on this issue? I think it's a crucial feature that needs to be added to the app, as nearly all other chess apps allow for OTB timed games.

arex1337:
+1 for this. 

> [22:12]
> i have realized that playing "OTB" on the lichess app is much better than playing OTB with a real board and pieces
> 
> [22:13]  
> it's much more convenient to carry and play on a phone, or preferably a tablet
> 
> [22:15]  
> i played on my tablet with some colleagues the other night, and it was awesome to be able to share the PGN by email with colleagues (and to import it to a study on lichess) (edited)
> 
> [22:18]  
> two things that could be better: 1) we used the lichess app clock functionality on a separate phone. a built-in clock in OTB mode would require 1 less device. we wouldn't have to hit the clock after we move. clock information could also be exported with the game in the future.

freefal:
https://github.com/veloce/lichobile/pull/665

PR #665, opened 2017-10-09
Add Option to Add Clock to OTB Games

freefal:
All the clock types supported in the `clock` module are supported for OTB games

Adds the feature requested in lichobile's oldest open issue here: https://github.com/veloce/lichobile/issues/99


veloce:
I see some lint errors. Please run `npm run lint` and fix all reported errors. Thanks!

veloce:
Thanks for this work! A couple of remarks after testing:

- clock could be bigger and displayed like in online games: aligned to the right of user name
- top clock should be rotated by 180deg, like pieces
- click sound like in standalone clock is not necessary: keep only move sound
- time is not saved if app is killed (and not paused): you should listen to cordova `pause` event and `window.onunload` too
- lots of debugging `console.log` statements are still there, please remove them.

freefal:
- Ok, I will do that
   - Good point. I'll rotate the name and the clock by 180 degrees.
   - Good point. I will have the clock code use a callback so we can make a
   sound in the clock module and be silent in the otb module
   - Thanks for pointing this out. I was listening for `remove` but not
   `pause` or `window.onunload`. I'll stop the clock and save its state on
   both events
   - Will delete. Sorry for sloppily leaving those in.


On Wed, Oct 11, 2017 at 3:36 AM, Vincent Velociter <notifications@github.com

src/ui/shared/clock/clocks/BronsteinClock.ts
```diff
@@ -18,6 +18,8 @@ export default function BronsteinClock(time: number, increment: number, onFlag:
     isRunning: false
   }
 
+  const id = Math.random()
```
veloce: `Math.random()` makes not a so good id. Prefer `performance.now()` which is guaranteed to be unique.

Also, I don't see why do you use the id?

src/ui/clock/clockSet.ts
```diff
@@ -1,50 +1,62 @@
...
+  return HandicapIncClock(time, increment, time, increment, onFlag, true)
 }
 
 export default {
+  none: () => null,
+
   simple: () => SimpleClock(
-    Number(settings.clock.simple.time()) * MINUTE_MILLIS
+    Number(settings.clock.simple.time()) * MINUTE_MILLIS,
+    () => {}
```
veloce: you can use `utils.noop` instead

src/ui/otb/clockSet.ts
```diff
@@ -0,0 +1,74 @@
+import settings from '../../settings'
+import HandicapIncClock from '../shared/clock/clocks/HandicapIncClock'
+import DelayClock from '../shared/clock/clocks/DelayClock'
+import BronsteinClock from '../shared/clock/clocks/BronsteinClock'
+import HourglassClock from '../shared/clock/clocks/HourglassClock'
+import StageClock from '../shared/clock/clocks/StageClock'
+
+
+const MILLIS = 1000
```
veloce: Should be exported in `../shared/clock`

src/ui/otb/OtbRound.ts
```diff
@@ -98,6 +102,19 @@ export default class OtbRound implements OtbRoundInterface, PromotingInterface {
       this.replay.init(variant, initialFen, situations, ply)
     }
 
+    if (data.offlineClock) {
+      const clockType = data.offlineClock.clockType as ClockType
```
veloce: casting shouldn't be necessary here

src/ui/otb/OtbRound.ts
```diff
@@ -98,6 +102,19 @@ export default class OtbRound implements OtbRoundInterface, PromotingInterface {
...
+      this.clock = clockSet[clockType](this.onFlag)
+      this.clock.setState(data.offlineClock)
+    }
+    else {
```
veloce: I don't understand this block. If there is no `data.offlineClock`, you shouldn't instantiate clock.

src/ui/otb/OtbRound.ts
```diff
@@ -115,6 +132,9 @@ export default class OtbRound implements OtbRoundInterface, PromotingInterface {
       payload.fen = setupFen
     }
 
+    const clockType = settings.otb.clock.clockType() as ClockType
```
veloce: I'd prefer have the clockType as a parameter of `startNewGame` instead of fetching localstorage.
This way you don't have to cast. And if arg is undefined it means it's a game without clock.

src/ui/otb/clockSet.ts
veloce: Should be imported from `../shared/clock`

src/ui/otb/newOtbGame.tsx
```diff
@@ -33,11 +37,30 @@ export default {
       isOpen(false)
     }
 
+    function addStage () {
```
veloce: Can't we reuse those function that must be defined in `../shared/clock`?

src/ui/otb/newOtbGame.tsx
```diff
@@ -115,3 +142,142 @@ export default {
     return null
   }
 }
+
+const clockSettingsView = {
```
veloce: Same here: I'm sure we can reuse settings view from `../shared/clock`

src/ui/shared/offlineRound/view.tsx
```diff
@@ -22,7 +24,7 @@ function getChecksCount(ctrl: OfflineRoundInterface, color: Color) {
     return 0
 }
 
-export function renderAntagonist(ctrl: OfflineRoundInterface, content: Mithril.Children, material: Material, position: Position, isPortrait: boolean, otbFlip?: boolean, customPieceTheme?: string) {
+export function renderAntagonist(ctrl: OfflineRoundInterface, content: Mithril.Children, material: Material, position: Position, isPortrait: boolean, otbFlip?: boolean, customPieceTheme?: string, clock?: IChessClock | null) {
```
veloce: Strange signature. You don't need the extra signification of `null` vs `undefined` here I suppose. So it should be `clock?: IChessClock`

src/ui/otb/OtbRound.ts
freefal: You are right.

src/ui/otb/OtbRound.ts
freefal: Agreed. Removed.

src/ui/clock/clockSet.ts
freefal: Done.

src/ui/otb/clockSet.ts
freefal: Agreed. Done.

freefal: Agreed. Done.

src/ui/shared/offlineRound/view.tsx
freefal: Agreed. At some point I was just trying to get things working and created this wacky signature. It's no longer necessary and has been adjusted to your suggested signature.

src/ui/otb/newOtbGame.tsx
freefal: Yes, good idea. Done.

src/ui/otb/OtbRound.ts
freefal: Makes sense. Done.

freefal:
Once merged,this branch will also resolve https://github.com/veloce/lichobile/issues/669

src/ui/otb/OtbRound.ts
```diff
@@ -98,6 +102,15 @@ export default class OtbRound implements OtbRoundInterface, PromotingInterface {
       this.replay.init(variant, initialFen, situations, ply)
     }
 
+    if (data.offlineClock) {
+      const clockType = data.offlineClock.clockType
+      this.clock = clockSet[clockType](this.onFlag)
+      this.clock.setState(data.offlineClock)
+    }
+
+    if (this.clock)
```
veloce: useless line

PR #666, opened 2017-10-10
Fix Tournament Conditions Coloring

freefal:


veloce:
Looks like a BC break. I haven't see any pb with condition coloring though.

Issue #604, opened 2017-06-03
OTB flip board


ornicar:
from email:

> In the Lichess App for iOS when using the " Over the Board" feature it does not allow you to Flip the board. So if you want to play Black versus a White opponent it does allow you to do it unless you physically change positions. Hope this can a fixed in a future upgrade. Thanks.

shubik22:
@veloce Hey, I'd be happy to take a shot at this if it's still a desired feature.

veloce:
Sure, go for it!

PR #667, opened 2017-10-11
Add ability to flip board after turn in OTB mode

shubik22:
Addresses https://github.com/veloce/lichobile/issues/604.  My first time contributing to the app so any feedback would be much appreciated -- thanks!

### DIFF
--- a/src/chessground/Chessground.ts
+++ b/src/chessground/Chessground.ts
@@ -125,9 +125,9 @@ export default class Chessground {
     anim(board.toggleOrientation, this)
   }
 
-  setOtbMode(mode: 'flip' | 'facing'): void {
+  setOtbFlipMode(mode: cg.OtbFlipMode): void {
     anim(state => {
-      state.otbMode = mode
+      state.otbFlipMode = mode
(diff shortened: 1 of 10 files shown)


veloce:
Thanks for this. Will review asap.

veloce:
hi @shubik22 
So sorry for the delay (I had bug fixes and more urgent features to merge), but now I am available to review.

I think by looking at the code that you added an option to flip board after each move, which I think is a nice feature if you play side by side and you pass the phone between each move.

From what I understood of #604 , it was asked that you can actually choose which side of the board play white or black. So it's more a feature for playing on a table facing each other (as first OTB mode was thought).

Still I will test it to be sure that's actually what I understood from reading the code, and in any case I'll merge it, since it's a useful feature :)

veloce:
@shubik22 I tested your PR. It works well and it's a nice feature to have.

One thing, though: it's a bit weird to see the change of board orientation while the piece is moving. To fix this, we should either disable animations for this mode, or switch board orientation after piece animation is complete.

Issue #661, opened 2017-10-07
Home improvements


veloce:
Some ideas:

- featured tournaments
- tournament leaderboard
- more links to menu items
- reorganize space

fabio656:
To have "Activity" in home like the site :)

PR #707, opened 2017-11-15
Home Screen Improvements

freefal:
Addresses https://github.com/veloce/lichobile/issues/661


freefal:
This is what the new home screen currently looks like. 
image

src/styl/home.styl
```diff
@@ -1,9 +1,39 @@
 .home
-  padding 10px
+  padding-top 5px
+  >*
```
veloce: I'd try to avoid that, by principle. '*' in css rule is highly inefficient.

src/styl/common.styl
```diff
@@ -125,7 +125,7 @@ span.fa
     margin-right 10px
   &:last-child
     border-bottom none
-  &.odd
+  &.even
```
veloce: What's the goal of this? I'm not sure but it may have an impact on the rest of the screens.

veloce:
Light and transparent themes are not handled (quick buttons and featured tournaments list).

src/styl/common.styl
freefal: If you switch it, you'll see that it's hard to see where the background ends and the first item begins. By making the first item (item 0) contrast with the background, there is a clearer delineation of where the list starts. I just ran around the whole app and made sure this worked everywhere it was used.
I also tried to fix some stuff that looked obviously wrong as I was going. (e.g., the tabs in the analysis view were colored dark)

src/styl/home.styl
freefal: Done.

freefal:
Fixed light theme. Not sure what the transparent theme is. I see references to it in the code base but no menu item to enable it. Sure I'm missing something obvious.

freefal:
Fixed custom themes.

src/styl/analyse.styl
```diff
@@ -291,7 +291,7 @@ comment
     color headerBgDarkColor
 
   .analyse-tabs
-    color darken(headerBgDarkColor, 5%)
+    background-color bglight
```
veloce: this change is certainly a mistake. it breaks analysis tabs style.

src/styl/custom_theme.styl
```diff
@@ -135,6 +135,17 @@
...
       padding-left 10px
 
+  .home
+    >.homeCreate
+      >.newGame-pools
+        >.newGame-pool
+          background-color transpWhite005
+          &.active
+            background-color transpWhite01
+  .ceval-pv
```
veloce: I don't get the need for this change. Please only change home screen style, since it's a PR that is home related

src/styl/light_theme.styl
```diff
@@ -162,3 +162,16 @@ bglightOdd = darken(themeBgLight, 5%)
...
+  .homeCreate
+    >.newGame-pools
+      >.newGame-pool
+        background-color darken(bglight, 2%)
+        border 1px solid #3d3d3d
+        color inherit
+        &.active
+          background-color darken(bglight, 8%)
+
+  .ceval-pv
```
veloce: this change breaks analysis style in light theme. please remove

PR #709, opened 2017-11-18
Fix formatting of chromium browser command

freefal:


TheCodeArtist:
The actual command that worked on my ubuntu 16.04 PC was `chromium-browser` instead of `chromium`.

 Is that usual or the exception? Should we update the command?...

ornicar:
It depends on the OS. I think sticking to `chromium` is fine.

PR #723, opened 2017-11-30
Add rapid time control

freefal:
I am still testing this so please do not merge yet, but I thought it would be useful to create a PR for this so you could see where I'm heading.


ornicar:
Cool, that's not much change.

PR #724, opened 2017-11-30
Offline puzzles

freefal:
The idea of this PR is fully replace the current online puzzle system with an offline puzzle system. The difference is that the offline system maintains an array of 25 puzzles (`unsolvedPuzzles` in settings) as well as a list of puzzles that have been attempted but not yet reported to the server (`solvedPuzzles in settings).

The idea is that if the user is offline or has a spotty connection, he can solve multiple puzzles while offline. Each time he attempts a puzzle while offline, that puzzle is removed from `unsolvedPuzzles` and the attempt is recorded in `solvedPuzzles`. After these changes to the two arrays are complete, the app attempts a sync with the server, whereby new puzzles are requested and the outcome of attempted puzzles is reported to the server. A sync is also attempted whenever the puzzle page is loaded.

In this way, the app becomes tolerant of being offline when loading a puzzle and when solving a puzzle such that puzzles can be played seamlessly on a spotty connection or no connection at all (up to 25 puzzles).


src/settings.ts
```diff
@@ -428,5 +430,14 @@ export default {
 
   importer: {
     analyse: localstorageprop('importer.analyse', false)
+  },
+
+  training: {
+    puzzleBufferLen: 50,
+    ratingDiffThreshold: 100,
+    unsolvedPuzzles: localstorageprop<OfflinePuzzle[]>('training.unsolvedPuzzles', []),
```
veloce: I'd put the last 4 props in a `training/state.ts` file instead, since it's not really settings but state.

src/ui/training/TrainingCtrl.ts
```diff
@@ -49,6 +50,11 @@ export default class TrainingCtrl implements PromotingInterface {
   }
 
   public init = (cfg: PuzzleData) => {
+    const user = settings.training.user()
```
veloce: Inject it in the constructor instead of having a hard dependency on settings.

src/ui/training/TrainingCtrl.ts
```diff
@@ -165,18 +171,23 @@ export default class TrainingCtrl implements PromotingInterface {
...
-    .then(cfg => {
+  public newPuzzle = () => {
+    const onSuccess = (cfg: PuzzleData) => {
       this.init(cfg)
-    })
-    .then(this.onXhrSuccess)
-    .catch(this.onXhrError)
+      this.onPuzzleLoad
+    }
+    loadOfflinePuzzle().then(onSuccess, puzzleLoadFailure)
```
veloce: Don't use `promise.then(success, failure)` form. It's a bad practice in general concerning promises, and in the rest of the app, we always use this form:
```
promise
.then(success)
.catch(failure)
```

src/ui/training/index.ts
```diff
@@ -7,22 +7,21 @@ import { emptyFen } from '../../utils/fen'
...
+import { State } from './interfaces'
+import { PuzzleData } from '../../lichess/interfaces/training'
 
 interface Attrs {
   id?: string
   initFen?: string
   initColor?: Color
 }
 
-interface State {
```
veloce: Leave interface State here, it's done like this in the rest of the app. It's more convenient to see what is the component interface directly where it's defined.

src/ui/training/index.ts
```diff
@@ -43,12 +42,12 @@ export default {
...
-      .then(cfg => {
+      const onSuccess = (cfg: PuzzleData) => {
         this.ctrl = new TrainingCtrl(cfg)
         cachedState.ctrl = this.ctrl
-      })
-      .catch(handleXhrError)
+      }
+      const afterSync = () => loadOfflinePuzzle().then(onSuccess, puzzleLoadFailure)
+      syncPuzzles().then(afterSync, afterSync)
     }
```
veloce: You can rewrite it like (betters style, to avoid `promise.then(succ, fail)`)
```
syncPuzzles()
.then(loadOfflinePuzzle)
.then((cfg: PuzzleData) => {
  this.ctrl = new TrainingCtrl(cfg)
  cachedState.ctrl = this.ctrl
})
.catch(puzzleLoadFailure)
```

src/ui/training/interfaces.ts
```diff
@@ -23,3 +28,7 @@ export interface VM {
   canViewSolution: boolean
   resultSent: boolean
 }
+
+export interface State {
```
veloce: remove (cf other comment)

src/ui/training/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  const user = settings.training.user()
+  const curRating = user ? user.rating : null
+  const discardedPuzzles: OfflinePuzzle[] = []
+  // Cull puzzles that were retreived when the user's rating was significantly different than it is now
+  if (curRating)
+    unsolved(unsolved().reduce((acc: OfflinePuzzle[], cur: OfflinePuzzle) => {
+      const ratingDiff = (curRating && cur.userRating) ? Math.abs(curRating - cur.userRating) : 0
+      if (ratingDiff < settings.training.ratingDiffThreshold) {
+        discardedPuzzles.push(cur)
+        acc.push(cur)
```
veloce: You should not mutate `acc` directly. Instead, return new `acc` like: `return acc.concat([cur])`

src/ui/training/utils.ts
```diff
@@ -0,0 +1,67 @@
+import * as xhr from './xhr'
+import { PuzzleSyncData, PuzzleData } from '../../lichess/interfaces/training'
+import router from '../../router'
+import settings from '../../settings'
+import { OfflinePuzzle } from './interfaces'
+
+export function syncPuzzles() {
+  const unsolved = settings.training.unsolvedPuzzles
```
veloce: You should define a const for `unsolved()` to avoid calling it multiple times (calling it requires access to filesystem).

src/ui/training/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+      })))
+      settings.training.user(syncData.user)
+      return true
+    }).catch(() => {
+      // If loading new puzzles fails, add back the discarded ones to avoid potentially running out of puzzles
+      unsolved(unsolved().concat(discardedPuzzles))
+      return false
+    })
+  }
+  if (settings.training.solvedPuzzles().length) {
```
veloce: Put `solvedPuzzles()` in a const to avoid calling it multiple times.

src/ui/training/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    xhr.solvePuzzles(settings.training.solvedPuzzles()).then(() => settings.training.solvedPuzzles([]), () => {})
```
veloce: Please put `.then()` and `.catch()` in a new line if the line is long, and avoid the 2 arg style in `.then()`

src/ui/training/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    return null
+}
+
+export function loadOfflinePuzzle(): Promise<PuzzleData> {
+  const cfg = loadNextPuzzle()
+  if (cfg !== null) {
+    return Promise.resolve(cfg)
+  }
+  else {
+    return Promise.reject('No additional offline puzzles available. Go online to get another ${settings.training.puzzleBufferLen}')
```
veloce: Use backtick otherwise string interpolation won't work

src/ui/training/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+      return false
+    })
+  }
+  if (settings.training.solvedPuzzles().length) {
+    xhr.solvePuzzles(settings.training.solvedPuzzles()).then(() => settings.training.solvedPuzzles([]), () => {})
+  }
+  return puzzlesLoaded
+}
+
+function loadNextPuzzle() {
```
veloce: it seems that this extra function is not necessary: you can put all the logic in `loadOfflinePuzzle()`

src/ui/training/index.ts
freefal: Sorry to be dumb, but can you walk me through the logic on this one? In the original version, we call `loadOfflinePuzzle` regardless of whether the call to `syncPuzzles` is successful. In the proposed version, doesn't a failure on `syncPuzzles` result in `loadOfflinePuzzle` never being called? Perhaps instead we could do this:
```
syncPuzzles(database)
        .then(() => loadOfflinePuzzle(database))
        .catch(() => loadOfflinePuzzle(database))
        .then((cfg: PuzzleData) => {
          this.ctrl = new TrainingCtrl(cfg, database)
          cachedState.ctrl = this.ctrl
        })
        .catch(puzzleLoadFailure)
 ```

src/ui/training/utils.ts
freefal: I had to put it in a local variable declared via a `let` rather than a `const` in order to actually save any function calls.

veloce:
Nice changes, this is better. Now there is still to address the cache router state cache issue (and fix the small conflict with master).

Issue #719, opened 2017-11-29
Tournament creation is broken


ornicar:
- Open the tournament creation screen
- Keep default settings, create the tournament
- "Error" is displayed

ornicar:
quoting myself from slack:

veloce:
Fixed by #726

PR #726, opened 2017-11-30
Fix tournament creation by supplying the position as a FEN rather than ECO code

freefal:


src/utils/fen.ts
```diff
@@ -1,4 +1,5 @@
 export const emptyFen = '8/8/8/8/8/8/8/8 w - - 0 1'
+export const initialPositionFen = 'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1'
```
veloce: Please use instead the constant already exported in `src/lichess/variant` as `standardFen`

veloce:
Fixes #719

ornicar:
About a week

Issue #739, opened 2017-12-16
autolink PMs


ornicar:
> Will it ever be possible to open links in PM in the app?
-- MeneerMandje

PR #742, opened 2017-12-20
Linkify Private Messages

freefal:
Closes https://github.com/veloce/lichobile/issues/739

Links are <a> tags with the target set to `_system`. The user then has to confirm that they wish to open the link in the app. Once they do, the link falls through all the same deep links logic we've already established.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -297,14 +297,14 @@ const linkPattern = /(^|[\s\n]|<[A-Za-z]*\/?>)((?:(?:https?):\/\/|lichess\.org\/
 function linkReplace(_: string, before: string, url: string) {
   const fullUrl = url.indexOf('http') === 0 ? url : 'https://' + url;
   const minUrl = url.replace(/^(?:https:\/\/)?(.+)$/, '$1');
-  return before + '<a target="_blank" rel="nofollow" href="' + fullUrl + '">' + minUrl + '</a>';
+  return before + '<a href="#" onClick="window.open(\'' + fullUrl + '\', \'_system\')">' + minUrl + "</a>";
 }
 
 const userPattern = /(^|[^\w@#/])@([\w-]{2,})/g
 
 function userLinkReplace(orig: string, prefix: String, user: string) {
   if (user.length > 20) return orig;
-  return prefix + '<a href="' + globalConfig.apiEndPoint + '/@/' + user + '">@' + user + "</a>";
+  return prefix + '<a href="#" onClick="window.open(\'' + globalConfig.apiEndPoint + '/@/' + user + '\', \'_system\')">@' + user + "</a>";
 }
 
 function autoLink(html: string) {
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -304,7 +304,7 @@ const userPattern = /(^|[^\w@#/])@([\w-]{2,})/g
 
 function userLinkReplace(orig: string, prefix: String, user: string) {
   if (user.length > 20) return orig;
-  return prefix + '<a href="/@/' + user + '">@' + user + "</a>";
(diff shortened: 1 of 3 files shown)


src/utils/index.ts
```diff
@@ -283,3 +285,28 @@ export function shallowEqual(objA: OAny, objB: OAny): boolean {
 
   return true
 }
+
+export function linkify(text: string): string {
+  const escaped = escapeHtml(text);
```
veloce: Please run `npm run lint` and fix any lint error (semicolon).

src/utils/index.ts
```diff
@@ -283,3 +285,28 @@ export function shallowEqual(objA: OAny, objB: OAny): boolean {
...
+  const fullUrl = url.indexOf('http') === 0 ? url : 'https://' + url;
+  const minUrl = url.replace(/^(?:https:\/\/)?(.+)$/, '$1');
+  return before + '<a href="#" onClick="window.open(\'' + fullUrl + '\', \'_system\')">' + minUrl + "</a>";
+}
+
+const userPattern = /(^|[^\w@#/])@([\w-]{2,})/g
+
+function userLinkReplace(orig: string, prefix: String, user: string) {
+  if (user.length > 20) return orig;
+  return prefix + '<a href="#" onClick="window.open(\'' + globalConfig.apiEndPoint + '/@/' + user + '\', \'_system\')">@' + user + "</a>";
```
veloce: User link should not open an external window, but user profile in the application.

src/utils/index.ts
```diff
@@ -283,3 +285,28 @@ export function shallowEqual(objA: OAny, objB: OAny): boolean {
...
+  const linked = autoLink(escaped);
+  return linked;
+}
+
+const linkPattern = /(^|[\s\n]|<[A-Za-z]*\/?>)((?:(?:https?):\/\/|lichess\.org\/)[\-A-Z0-9+\u0026\u2019@#\/%?=()~_|!:,.;]*[\-A-Z0-9+\u0026@#\/%=~()_|])/gi;
+
+function linkReplace(_: string, before: string, url: string) {
+  const fullUrl = url.indexOf('http') === 0 ? url : 'https://' + url;
+  const minUrl = url.replace(/^(?:https:\/\/)?(.+)$/, '$1');
+  return before + '<a href="#" onClick="window.open(\'' + fullUrl + '\', \'_system\')">' + minUrl + "</a>";
```
veloce: String concatenation is hard to understand here. Please rewrite it so:
`<a href="#" onClick="window.open('${fullUrl}', '_blank')">${minUrl}</a>`

Notice the `_blank` instead of system: it opens a link in the in-app browser. This is consistent with other links in the app.

freefal: The issue with `_blank` is that it won't redirect the user to the appropriate section of the app if the app could handle the link. For example, if the user clicked on a link like `https://lichess.org/tournament/QecWLuj9` and we set the target to `_blank` it will open that website version of the tournament page in an inapp browser. By using the `_system` target, the user will be prompted to either open the link in a proper browser or the app. If they choose the app, it will fall through out normal deep link regex rules defined in `config.xml` and `deepLinks.ts` so if the app can handle the link the user is directed to the correct app component and if the app can't handle the link, the link is opened in the inapp browser.
This is all consistent with how clicking a lichess.org link outside the app would work. While it involves kicking the control flow out to the OS for a moment, I think it's the best approach as it doesn't involve reimplementing the rules we've defined in `config.xml` again.

src/utils/index.ts
freefal: Here again, the user will be prompted as to whether they wish to open the link in a browser or the app. However, here it probably does make sense to just go directly to the user page in the app.

freefal:
Fixed.

ornicar:
nice!

PR #743, opened 2017-12-20
Fix User TV

freefal:
@ornicar learned from a user that opening a link like `lichess.org/@/freefal/tv` in the app will just open the user profile page for the user named `tv`.
This fixes that issue.


src/deepLinks.ts
```diff
@@ -24,6 +24,7 @@ export default {
     universalLinks.subscribe('tv', () => router.set('/tv'))
     universalLinks.subscribe('tvChannel', (eventData: UniversalLinks.EventData) => router.set('/tv/' + eventData.path.split('/').pop()))
     universalLinks.subscribe('userVariantProfile', handleVariantProfile)
+    universalLinks.subscribe('userTV', (eventData: UniversalLinks.EventData) => { console.log(eventData.path.split('/')[2]); router.set('/@/' + eventData.path.split('/')[2] +'/tv')})
```
veloce: please remove console.log

freefal:
Fixed.

PR #748, opened 2018-01-04
Allow "@" in URLs for auto linking

niklasf:
Untested 😨 

Similar to https://github.com/ornicar/lila/commit/36d70d9c29dd839f7965d04fc7eea471a6c6f6b5


veloce:
Thanks, will test it asap.

Issue #716, opened 2017-11-27
handle /games/search being down


ornicar:
It's down until we find a solution for indexing that many games: https://lichess.org/games/search. Actually it looks like that on the app:

![](https://cdn.discordapp.com/attachments/352976626558042122/384710768089890827/2017-11-27_19.23.41.jpeg)

veloce:
That's a recent change that cause it, if I can't parse the response as json, I use the `responseBody` as text.
I will fix it globally on the app side to never show an error message that is text or html.

It would be better if the server would always return json, even when status is in error. Is there a way in play to add a global hook so errors are never returned as html, and instead return a simple json using request status: 
```
{
    "error": status text
}
```

that can be overriden on a specific route ?

TheCodeArtist:
If all this involves is **testing/parsing a JSON response and throwing a toast on Android**, i would like to fix this.

*(FYI am new to JS, picked-up JSON parsing while writing my first FireFox webextension [Deep thought Tabs](https://github.com/TheCodeArtist/deep-thought-tabs) last month.)*

I have setup the mobile app dev setup by following the [Readme](https://github.com/veloce/lichobile/blob/master/README.md).  
Currently stuck setting-up a local lila server as launching the mobile-app in chromium simply gave me a static unresponsive mobile-app home-page. (i used the contents of `env.json.example` as-is to create `env.json`)

Is this expected? Is there a public dev lila server i can point in my env.json to quickly start working on the mobile-app?

TheCodeArtist:
@veloce Wow! that was fast!

veloce:
Sorry, I fixed it before I saw you comment.

There is a public lila server, this is the one referenced in env.json.example :)
I should add a section to show how to use this server and CORS to launch the app in a browser, but I need to fix a couple of issue regarding cors before.
You can still start a dev browser without web security in the meantime.. the command is in the readme.

TheCodeArtist:
Thanks. i had followed the steps. But the UI was unresponsive.
![lichobile-chromium](https://user-images.githubusercontent.com/1897177/33277776-8cfee4ae-d3bf-11e7-915b-a737d06ea06d.jpg)

Turns out we need to enable "Device Mode" from the dev tools.
(dunno how i figured it out! 😕 )
![device-mode-chromium](https://user-images.githubusercontent.com/1897177/33278015-3840e402-d3c0-11e7-8d6b-2747d66d3275.jpg)

[...]

PR #755, opened 2018-01-08
Clarify how to run in browser.  I ran into this problem, and luckily …

dsjoerg:
…came across https://github.com/veloce/lichobile/issues/716 explaining how to resolve my problem.


veloce:
Thanks.

PR #769, opened 2018-02-20
Clicking on a game in a player's game list that is live should route the user to that game

freefal:
rather than the analysis board for that game

### DIFF
--- a/src/ui/user/games/gamesView.tsx
+++ b/src/ui/user/games/gamesView.tsx
@@ -8,6 +8,7 @@ import i18n from '../../../i18n'
 import session from '../../../session'
 import spinner from '../../../spinner'
 import GameItem from '../../shared/GameItem'
+import gameStatus from '../../../lichess/status'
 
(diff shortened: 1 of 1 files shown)


src/ui/user/games/gamesView.tsx
```diff
@@ -54,7 +55,7 @@ function onTap(ctrl: IUserGamesCtrl, e: Event) {
         const userColor: Color = g.players.white.userId === userId ? 'white' : 'black'
         positionsCache.set(g.id, { fen: g.fen, orientation: userColor })
         const mePlaying = session.getUserId() === userId
-        if (mePlaying) {
+        if (mePlaying || (g.source !== 'import' && g.status.id < gameStatus.ids.aborted)) {
```
veloce: You should use this function instead: 

https://github.com/veloce/lichobile/blob/master/src/lichess/game.ts#L29

which does the exact same check

veloce: ok, nevermind my comment, this game object is not compatible...

PR #787, opened 2018-03-21
Highlight current move in ongoing game analysis

bak1an:
The same way it is done on the desktop.

Before:
screenshot 2018-03-21 20 51 15

After:
screenshot 2018-03-21 20 51 36


veloce:
Nice! Thanks.

PR #791, opened 2018-03-31
Training: Add votes to puzzles

### LINKED ISSUE #660: Add vote to training screen

DanielShaulov:
Added vote count and ability to vote at the end of puzzles.

This fixes #660.
The functionality is correct, but it doesn't look too good - I couldn't make it align to the right.
I would really appreciate some help with the styling part.

### DIFF
--- a/src/ui/training/trainingView.ts
+++ b/src/ui/training/trainingView.ts
@@ -154,12 +154,45 @@ function renderViewSolution(ctrl: TrainingCtrl) {
   }, i18n('viewTheSolution')) : h('div.buttonPlaceholder', '')
 }
 
+function renderVoteControls(ctrl: TrainingCtrl) {
+  if (ctrl.vm.voted === true) {
+    return [
+      h('span.fa.fa-caret-up'),
+      h('p', ctrl.getVotes()),
+      h('a', {
+        oncreate: helper.ontap(ctrl.downvote),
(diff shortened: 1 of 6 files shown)


src/ui/training/trainingView.ts
```diff
@@ -154,20 +154,54 @@ function renderViewSolution(ctrl: TrainingCtrl) {
   }, i18n('viewTheSolution')) : h('div.buttonPlaceholder', '')
 }
 
+function renderVoteControls(ctrl: TrainingCtrl) {
+  if (ctrl.vm.voted === true) {
+    return [
+      h('span.fa.fa-caret-up'),
+      h('p', ctrl.getVotes()),
+      h('a', {
```
veloce: I'd prefer a `button` instead of a `a`. It's not a website, and there is not much point to have an anchor tag without `href`.

veloce:
It's hard to explain how to style, so I preferred to do it directly, hope you don't mind.
You can see e38ddad76 I just simplified the view and changed the style (also added checks to not show controls if not logged in). I changed the alignment because on some devices there is not much vertical space, so we can't the vote buttons on top and on bottom of the number. Instead I put them aside.

DanielShaulov:
I don't mind at all. Thanks a lot!

PR #792, opened 2018-03-31
Add score to the piece cemetery

DanielShaulov:
Hi,
I added the little score next to the piece cemetery like in the lichess website.
I again have some problems with the styling, the number is a bit lower than the pieces. The margin-top that I placed on the score seems to no work like it does on the piece.


veloce:
Thanks for this!

PR #795, opened 2018-04-05
Add Score to Game Popup Menu

freefal:
Utilizes API addition implemented here (https://github.com/ornicar/lila/commit/17a77815203be6e552aa7622e13a859285ce7518)


src/ui/shared/round/view/roundView.tsx
```diff
@@ -478,13 +498,19 @@ function renderGamePopup(ctrl: OnlineRound) {
...
     ctrl.vm.showingActions,
     ctrl.hideActions
   )
 }
 
+function renderButtonsAndScore(ctrl: OnlineRound): Mithril.Children {
+  const buttons = gameApi.playable(ctrl.data) ?
+    () => renderGameRunningActions(ctrl) :
+    () => renderGameEndedActions(ctrl)
+  return [buttons(), renderScore(ctrl)]
```
ornicar: why is `buttons` a function?

src/ui/shared/round/OnlineRound.ts
```diff
@@ -185,6 +188,14 @@ export default class OnlineRound implements OnlineRoundInterface {
   public showActions = () => {
     router.backbutton.stack.push(this.hideActions)
     this.vm.showingActions = true
+
+    const d = this.data
+    if (!d || !d.player.user || !d.opponent.user)
+      return
+    xhr.getCrosstable(d.player.user.id, d.opponent.user.id).then(s => {
```
veloce: It is unnecessary to request the crosstable each time the popup is opened. You should do it only once.

freefal: I thought it was worth requesting each time in case the game ended and the score changed, but in retrospect the cost/benefit of that decision doesn't make any sense. Fixed.

veloce: Then you could add a method `updateCrosstable()` and call it first time before showing popup, and once when the game has just ended (there is already a hook for that).

freefal:
Thanks. I just did this and committed it.

On Fri, Apr 6, 2018 at 10:42 AM, Vincent Velociter <notifications@github.com

src/http.ts
```diff
@@ -16,6 +16,7 @@ export interface RequestOpts {
   body?: any
   query?: Object
   headers?: StringMap
+  cache?: 'default' | 'no-store' | 'reload' | 'no-cache' | 'force-cache'
```
veloce: You can use `RequestCache` interface instead, which comes with the standard lib.

src/ui/shared/round/util.ts
```diff
@@ -12,3 +12,10 @@ export function lastPly(d: OnlineGameData): number {
 export function plyStep(d: OnlineGameData, ply: number): GameStep {
   return d.steps[ply - firstPly(d)]
 }
+
+export function getWhiteBlack(d: OnlineGameData): Array<User | undefined> {
```
veloce: I'd prefer a `getPlayer(d: OnlineGameData, color: Color): Player` function

freefal:
Should the function remain in util.js or should we move it to gameApi?

veloce:
gameApi sounds fine to me.

freefal:
...and it already exists there ;)

freefal:
I think it looks ok, but I remember that ornicar agreed to create this API
only on the condition that we didn't add clutter to the game screen.

What about adding this info (or the full crosstable as you had suggested)
to the small pop-up that is displayed when the user clicks on a player's
name?

On Wed, Apr 11, 2018 at 1:21 PM, Vincent Velociter <notifications@github.com

veloce:
@freefal yes I think we'll do that. In the user popup we can be more verbose. We can add the number of games, and say it's the lifetime score. Good idea.

PR #796, opened 2018-04-05
Add about page

freefal:
It probably makes sense to add a paragraph on the history of the mobile app, but I wasn't around so can't author it.

### DIFF
--- a/src/ui/helper/index.ts
+++ b/src/ui/helper/index.ts
@@ -365,3 +365,15 @@ export function findElByClassName(e: Event, className: string) {
     target : findParentBySelector(target, '.' + className)
 }
 
+export function externalLink(text: string, url: string): Mithril.Children {
+  return h('span', { class: 'link' }, [
+    h('a', { href: '#', onClick: 'window.open("' + url + '")' }, text),
+    h('span', { class: 'fa fa-external-link' })
+    ])
+}
+
+export function internalLink(text: string, route: string): Mithril.Children {
+  return h('span', { class: 'link' }, [
+    h('span', { oncreate: ontap(() => { router.set(route) }) }, text),
+  ])
+}
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -36,6 +36,7 @@ import inboxThread from './ui/inbox/thread'
 import inboxCompose from './ui/inbox/compose'
 import importer from './ui/importer'
 import search from './ui/search'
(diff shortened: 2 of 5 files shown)


src/ui/helper/index.ts
```diff
@@ -365,3 +365,15 @@ export function findElByClassName(e: Event, className: string) {
     target : findParentBySelector(target, '.' + className)
 }
 
+export function externalLink(text: string, url: string): Mithril.Children {
+  return h('span', { class: 'link' }, [
+    h('a', { href: '#', onClick: 'window.open("' + url + '")' }, text),
+    h('span', { class: 'fa fa-external-link' })
+    ])
```
veloce: indentation

src/ui/helper/index.ts
```diff
@@ -365,3 +365,15 @@ export function findElByClassName(e: Event, className: string) {
     target : findParentBySelector(target, '.' + className)
 }
 
+export function externalLink(text: string, url: string): Mithril.Children {
+  return h('span', { class: 'link' }, [
+    h('a', { href: '#', onClick: 'window.open("' + url + '")' }, text),
```
veloce: use a template string, and add the `_system` arg:
```h('a', { href:"#", onClick: `window.open('${url}', '_system')` }, text)```

File: src/ui/helper/index.ts [hunk 3]
```diff
@@ -365,3 +365,15 @@ export function findElByClassName(e: Event, className: string) {
...
+    h('span', { class: 'fa fa-external-link' })
```
veloce: don't add this, instead use the `external_link` class on the `a` element.
see: https://github.com/veloce/lichobile/blob/master/src/utils/html.ts#L31

veloce:
Maybe I can add a paragraph on the the mobile app. Also we can add links as in the website, and add the link to this repo too, since it's the about of the app.

src/ui/helper/index.ts
```diff
@@ -365,3 +365,15 @@ export function findElByClassName(e: Event, className: string) {
     target : findParentBySelector(target, '.' + className)
 }
 
+export function externalLink(text: string, url: string): Mithril.Children {
+  return h('span', { class: 'link' }, [
```
veloce: No need to have an `a` inside a `span`. Also no need for class `link`, so remove the span element.

PR #797, opened 2018-04-06
Make puzzle orientation explicit

freefal:
Without this, offline puzzles were always shown from white's perspective when moving to the analysis board.


veloce:
Thanks! It shows that ` 'analyse/:source/:id/:color'` is a useless route, I should remove it.

PR #804, opened 2018-04-10
Change the bookmark icon to change locally in round view

DanielShaulov:
After successfully toggling the bookmark in the server, there is no need for a separate request to update the gamedata just to toggle the UI. This is the way it is already done in the search and game list views.

Also, there seems to be some problem in the server - it takes some time for it to start returning "bookmarked = true"  in the gameData.  This causes the UI to not toggle at all, this PR works around this issue.


veloce:
Thanks

PR #806, opened 2018-04-10
Make hardware keys control media by default

freefal:
A pain point for me and (I suspect) others with the app is that sometimes the piece movement volume is too loud or too soft, but using the volume keys defaults to adjusting the ring volume. This PR changes the default volume controlled by the hardware keys to instead be the media volume.


veloce:
Sweet!

PR #827, opened 2018-05-07
Two-factor authentication

niklasf:
Here's an attempt to implement two factor authentication. It works nicely in Chrome.

There's a new translation key `invalidAuthenticationCode` in lila that I don't know how to get.

Another point worth checking with a real device would be if the numeric keypad is selected.


veloce:
App still uses old i18n system, that's why you can't get `invalidAuthenticationCode`. I suggest only writing in english for now.

I'll check it asap.

niklasf:
Ok, switched to plain English for now.

Issue #836, opened 2018-05-22
bug: empty lists top margin


apunko:
# Bug report

**Current behavior:**
The top margin is absence on empty lists.

![image](https://user-images.githubusercontent.com/10829386/40362225-922cd734-5dd4-11e8-8876-638b7b9d35e5.png)

**Expected behavior:**
Top margin. 'Oops! Nothing here.' under tab titles without interference.

**Steps to reproduce:**
Precondition: Have no challenges
1) Open 'Correspondence' from user menu
2) Challenges tab

The same behavior for Public Games.

**Environment information**
IOS 10.3.1
lichess v5.6.1

I have added https://github.com/veloce/lichobile/pull/837 to fix it.
I really like and enjoy lichess and want to make contributions sometimes.

veloce:
Fixed in master.

PR #837, opened 2018-05-22
issue-836: add top margin to empty lists

apunko:
fixes https://github.com/veloce/lichobile/issues/836


veloce:
Thanks for this. But actually I'd prefer text in empty lists to be vertically centered (it was the case before, #836 is a regression).

PR #895, opened 2018-08-20
[Untested] Fix game links with colors so that they correctly use the color attr

freefal:


freefal:
This has now been tested and should be good to pull when you're ready.

Issue #898, opened 2018-09-09
How to update translations


erral:
Until #639 is solved, how can I update a translation here? Should I provide a PR for the relevant JSON file?

Thanks,

Issue #639, opened 2017-08-22
Switch to crowdin translations


veloce:

jas14:
Hey @veloce, can I pitch in here?

veloce:
Hi @jas14, of course! That's not a trivial change tough. You should contact @ornicar as well because it will probably require some synchro work with the main lichess repo (ornicar/lila).

jas14:
Hey @ornicar, any thoughts on how to get this started? Should/can we create a new project in the lichess Crowdin org, piggyback off the existing project, or would you rather thid be completely separate?

@veloce, I'm thinking of pulling in a JS i18n library instead of the current manual solution, probably i18next. Any opinions or thoughts on that?

jas14:
ping @ornicar

veloce:
@jas14 as for the i18n library I'd rather not use it and stick to a manual solution that you can update of course. 

You can perhaps take the lila module and use it in the app:
https://github.com/ornicar/lila/blob/master/ui/site/src/standalones/trans.js

Only a couple of functions, easy to understand, and all are useful. Much less code than a whole library.

ornicar:
I can now create a new open source project but "lichobile" is already taken - by you, I think https://crowdin.com/project/lichobile

jas14:
@ornicar I did start that project to test it out, if we go with this option I'll invite you to manage it too.

But, if we do start this separate project, the translations and original strings will diverge from the lila project. Are we concerned about divergence?

The alternative is for me to write a script that downloads the translated XML files from the lila project and convert those into JSON (dealing with plurals somehow). The downside is that we won't be able to handle any strings that are in the lichobile but not lila - there's already a blob of those in app.js.

If nobody objects, I'll submit a PR for the diverging option.

superuser-does:
Hi @jas14, I see one of our contributors found your Crowdin project on their own accord. Most of the strings are currently duplicated between the site and your project.

So to prevent duplication of effort for translators, I think it would be best to continue as we currently have and put them all into the main file (`site.xml`) if they affect both. We can split off any strings that are used only for lichobile (there are none that I'm aware of personally) into a separate file but under the same project. I think translators can reasonably expect it as this is the _official_ Lichess mobile app.

Thoughts @veloce and @ornicar (cc @niklasf)?

niklasf:
Both options (seperate project or same project) seem reasonable.

[...]

PR #901, opened 2018-09-18
Update & improve italian translation

renyhp:
Most of the changes proposed here are consistent to what is already on the site via Crowdin.


jas14:
@renyhp , how have you been dealing with editing the single-line JSON translation blob? Asking as I'm trying to localize a few more strings.

renyhp:
@jas14 it's been quite some time ago so I can't remember the exact way I have done it, probably used Visual Studio which formatted it automatically in a readable way. Anyway googling "JSON formatter" gives you a lot of possibilities, I have just found [this website](https://jsonformatter.curiousconcept.com/) which will surely do the job. As for maybe putting the JSON in place in a single line, probably just searching for tabs and newlines and replacing with nothing will do the trick.

Anyway I don't see much interest in this kind of PR (which is quite reasonable since it seems there are plans to switch to the Crowdin translations), so probably it's not worth the effort. If you want to contribute to translation I would suggest you do that [directly in Crowdin](https://crowdin.com/project/lichess).

jas14:
Any pointers to conversation(s) on switching to Crowdin? Seems like a great solution, I'd love to follow along on progress (and/or contribute)!

renyhp:
There's no actual conversation going on but the relevant issues are #639 and #898.


-------- Messaggio originale --------
Oggetto: Re: [veloce/lichobile] Update & improve italian translation (#901)
Da: Joe Stein
A: veloce/lichobile
CC: renyhp ,Mention



Any pointers to conversation(s) on switching to Crowdin? Seems like a great solution, I'd love to follow along on progress (and/or contribute)!

PR #959, opened 2019-03-10
Fix horizontal scrolling issue in TabView component

Arth-ur:
There was an issue with horizontal scrolling inside TabView component in the Tournaments List and Correspondence views:

bug

Here is a fix that wrap the tabs view inside a div with hidden horizontal overflow:

fix

### DIFF
--- a/src/styl/tabs.styl
+++ b/src/styl/tabs.styl
@@ -34,6 +34,9 @@
 
 
 // tab content
+.tabs-view-wrapper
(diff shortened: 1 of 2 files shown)


veloce:
Thanks. I fixed it here: https://github.com/veloce/lichobile/commit/ae0a47b6c7af57420fda94facf1e164a7b11836c (in a simpler way).

Issue #876, opened 2018-07-13
Tactics training repeats already solved positions


gcp:
- bug report

Open Android app, go to training. Observe that the puzzle is a familiar one: it has been solved on the lichess.org website the week before. Same for all following puzzles.

**Expected behavior:**

If an internet connection is present, puzzles that were already solved should not be repeated.

(If no connection is present, it would be nice if the batch of 50 that is downloaded was seeded in a different way from the website, to make it very unlikely that one gets dupes. Might also fix this bug.)

lichess seems to know the puzzle has already been solved, as my rating does not change.

**Environment information**

- Android 8.1
- Pixel 2 XL
- App version 6.0.4

veloce:
I acknowledge the problem, but this is also server side to me. That'd be nice to open as well a ticket on ornical/lila as well, pointing this one as reference. Server issue will be a better place to discuss the solution to the problem.

gcp:
Can you file that? I'm not clear what the underlying issue is? No API to probe whether positions are already solved? Or wanting the ability to "seed" the selection of positions that one gets?

veloce:
Yes I'm going to report the issue.

KommX:
@veloce Is there anything we could do to improve the situation with the puzzles?

supertorpe:
I have implemented a workaround that consists of a button to resynchronize the puzzles. For it:
1) I added a method "database.clean": return asyncStorage.removeItem (`offlinePuzzles. $ {UserId}`)
2) new method "TrainingCtrl.resync" that invokes database.clean and syncAndLoadNewPuzzle
3) new "fa-refresh" button in TrainingView
In summary: when the resynchronization button is pressed, the local puzzle cache is emptied and a new one is loaded from the server.

jakub-g:
As a kind of workaround for this issue, it would be handy to be able to
- skip a puzzle, or
- have a "random puzzle" button, or
- open a puzzle by id.

In the browser I can just change the URL to some random ID, but in mobile app there's no such possibility.

veloce:
No more relevant with new puzzle api it seems.

PR #960, opened 2019-03-13
issue #876: workaround - new button to resync puzzles

supertorpe:
Related issues https://github.com/veloce/lichobile/issues/876 and https://github.com/ornicar/lila/issues/4409

The synchronization  problem between the puzzles between the website and the app is that the app maintains an offline cache of puzzles. 
The solution I propose is not the most elegant, but after so many months of suffering the incidence, I think it's worth it until we get something better. It tries to add a button to invalidate the local cache and re-request a new batch of puzzles to the server.

You can see the new "resync" button to the right of the analysis button:

issue#876-workaround

### DIFF
--- a/src/ui/training/TrainingCtrl.ts
+++ b/src/ui/training/TrainingCtrl.ts
@@ -121,6 +121,26 @@ export default class TrainingCtrl implements PromotingInterface {
     return true
   }
 
+  public resync = () => {
+    const user = session.get()
+    if (user) {
+      const onSuccess = (cfg: PuzzleData) => {
+        this.vm.loading = false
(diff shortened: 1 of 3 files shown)


veloce:
Hi. Thanks for submitting this. This is indeed a known issue and to properly fix it we should change entirely the way the server sends its puzzle to the app to not mix puzzles cached in the app and puzzles played on the website. Such change is not planned right now, and afaik not even in the roadmap. So your solution is probably a good workaround.

DarthFlappy:
As far as I can tell, this update has not been put on the iOS app.

PR #962, opened 2019-03-17
Upgrade cordova-android from 7.1.0 to 8.0.0 to fix Gradle errors

freefal:


veloce:
@freefal what kind of errors does it fix? I guess it may also fix the build issue we had with android ndk, so we could try to restore it.

PR #963, opened 2019-03-20
Fix analysis scrolling

freefal:
Reapply patch that @veloce originally wrote to fix the analysis scrolling bug, which has reappeared in Chrome 74.


MarcoLaube:
Can you release an HOTFIX for this. I think it is really a major problem for people using the analysis section, which it is use in so many way inside lichess app (simple game analysis, studies, ect)

PR #965, opened 2019-03-25
Fix infinite scrolling for user games screen. 'onscroll' event was previously not firing

freefal:


veloce:
Fixed here https://github.com/veloce/lichobile/commit/ba106b12d8d538eb7adcf90ea391f0739d37985d and here https://github.com/veloce/lichobile/commit/e6fb0a480a4fc529dd078a64a3fc4f203feb9198

PR #980, opened 2019-05-10
Fix package.json and upgrade packages

jas14:
`com.rjfun.cordova.plugin.lowlatencyaudio` seems to be unused / duplicated in `cordova-plugin-nativeaudio`, but was messing with npm (which was reporting that the former was not installed).

This PR:
- fixes the above issue
- upgrades several packages found by `npm audit` to be causing security issues
- replaces the deprecated `gulp-util` with `fancy-log` and `chalk`

Tested with `npm run build-stage`.


veloce:
Thanks for this. I'll test asap on my computer. I believe `com.rjfun.cordova.plugin.lowlatencyaudio` is added automatically by `cordova run` command, so you can't really remove it.

PR #983, opened 2019-05-27
Add the 'letter' piece set from the lichess website.

definitelynobody:
A simple piece set that is currently used on the website and missing in the mobile app.


veloce:
Thanks!

PR #992, opened 2019-06-10
UI/UX improvements

veloce:


ornicar:
Wow, is that a visual update?

veloce:
My goal is to provide a better integration with the website, now that it has a responsive design, and since we already have way to provide authenticated sessions in the in-app browser.
It will work well for features such as profile edition and forums so I'll do that. Broadcast perhaps too, but it would be better to integrate them directly in the app since it has already read-only studies.

Home refactoring will allow to remove some menu items (play with a friend, with machine) and poorly designed correspondence screen. So the app home will look more like the website home (2 tabs for quick setup and corres.).

PR #995, opened 2019-06-14
Upgrade Stockfish plugin to fix build issues

jas14:
Upgraded `cordova-plugin-stockfish` to the most recent commit (https://github.com/veloce/cordova-plugin-stockfish/commit/7cc6722888bf482584205ca753c288826786c73b) to fix the Android NDK build issues. See https://github.com/veloce/cordova-plugin-stockfish/pull/14.


veloce:
Thanks. Is it possible to avoid changing the whole indentation of `package.json`?

jas14:
Yeah, I'll take care of that. I imagine this happened because, per the requirements in the README, I'm using Cordova 7.x (7.1.0 specifically), which was released before https://github.com/apache/cordova-lib/pull/606 . Any reason not to upgrade to the latest release, 9.0.0?

veloce:
Oh no, of course, you should use the latest cordova. I'll update the README.

Issue #1009, opened 2019-07-26
Back button issue


HBiSoft:
**I'm submitting a bug report**

**Current behavior:**

Whenever I press the back button, the application is closed (no matter where I am inside the application).

**Expected behavior:**

When I press the back button it should go to the previous page. When I'm busy playing a game and I accidentally press the back button it should not exit out of the application.

**Steps to reproduce:**

Open a tournament, for example, then press the back button. Like I mentioned above, it doesn't matter on what page I am inside of the application.

**Environment information**

- Samsung J7 Pro
- Android 9
- Lichess 6.1.4

---
I think it's also worth mentioning - Yesterday I updated the app for the first time in about 4 months. This wasn't an issue before I updated.

veloce:
I can reproduce this bug too (even though it does not happen all the time). Currently investigating.

veloce:
The cause is probably: https://github.com/apache/cordova-android/issues/687

freefal:
PR that I believe resolves this issue here: https://github.com/veloce/lichobile/pull/1022

HBiSoft:
Fixed in the new update.

Issue #985, opened 2019-06-01
[Bug] Back button terminates game immediatelly


MateMachine:
Lichess version: v6.1.3
Smartphone: Samsung Galaxy S8
Android version: 9
Rooted/CFW: No
-----------------------
Hello,

When I press the back button while I'm playing a match, the app gets terminated without any warning.
After I start the app again, I'm on the main menue.
The same happens when:
- I'm in the analysis board.
- I browse through settings
- play offline against the CPU

I'm sure in the previous version there was not such a problem.
Generally I was always able to go back to the previous matches or menues.

P.S. I'm not logged in - I use it only anonymously.

Can you check, please?

freefal:
PR that I believe resolves this issue here: https://github.com/veloce/lichobile/pull/1022

PR #1022, opened 2019-08-19
Fix back button on Android

freefal:
This comment (https://github.com/apache/cordova-plugin-splashscreen/pull/225#issuecomment-516405842) suggested a fix to the back button issue that was raised here (https://github.com/veloce/lichobile/issues/1009) and here (https://github.com/veloce/lichobile/issues/985)

I've implemented the suggested fix and it works. I confirmed the issue existed on my phone and disappeared once I added the extra parameter to the `<widget>` tag.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='utf-8'?>
-<widget android-versionCode="6001049" id="org.lichess.mobileapp" version="6.1.4" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
+<widget android-versionCode="6001049" id="org.lichess.mobileapp" version="6.1.4" xmlns="http://www.w3.org/ns/widgets" xmlns:android="http://schemas.android.com/apk/res/android" xmlns:cdv="http://cordova.apache.org/ns/1.0">
     <name>lichess</name>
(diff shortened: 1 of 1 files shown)


veloce:
The bug was caused by the cordova-plugin-splashscreen it seems. Could you also try to update it and see if it still works with your fix?

freefal:
Yes, will give it a shot tonight.

To be clear, the comment (
https://github.com/apache/cordova-plugin-splashscreen/pull/225#issuecomment-516405714)
I pointed to was on a still open PR that purports to fix the issue.

I will update cordova-plugin-splashscren to the latest and confirm
compatibility with my PR, but so long as
https://github.com/apache/cordova-plugin-splashscreen/pull/225 remains
open, I anticipate my PR will still be necessary.

veloce:
I trust you that the fix work and I'll merge as is. Will try to update the splashscreen later. Thanks.

freefal:
Sounds good. I upgraded splash screen and had no conflicts, but then I couldn’t reliably break the back button again by downgrading or removing the change I made, which I ran out of time to investigate. Can say with confidence that upgrading splashscreen shouldn’t conflict with config.xml change. Nor do I think the upgrade fixes the issue, as the PR is still open on github.

Just wanted to fix/break the back button by adding/removing my change and for some reason was having trouble reliably reproducing. Not sure why.

PR #1036, opened 2019-09-20
New i18n fetch script

jas14:
Part of #639 .

To run: `cd scripts; npm install; npm run i18n-fetch [locale1 [locale2 [...]]]`.

This script:
1. downloads the translations from Crowdin,
2. transforms them from XML to a flat JSON file,
3. merges them into the old JSON files, preserving any strings that are not present in the Crowdin translations.

Notably, the translation JSON files *no longer contain keys with fallback English values*; instead, they contain only the keys that have been translated. As such, users in locales with untranslated strings will see the keys instead of the English values.


veloce:
Very nice!
So do we use a separate crowdin project, or the same as lila?
What is the next step now?

Many thanks for that, I'll review an merge asap.

jas14:
For strings that we share with lila, this PR lets us use their translations from crowdin! 😃

For strings that are only on lichobile, there is no solution right now. I think the best option will be to open a separate crowdin project. That will require some modification of the script, but I'm happy to do that in a second PR so you don't have to review too much here.

veloce:
Ok. I saw in the diff that this adds a lot of new strings to the JSON files.
So the next step is probably to identify parts of the app where we can use those new strings (if they fit in the smaller space of the mobile app)

src/i18n.ts
```diff
@@ -9,6 +9,9 @@ let numberFormat: Intl.NumberFormat = new Intl.NumberFormat()
 
 export default function i18n(key: string, ...args: Array<string | number>): string {
   let str: string = messages[key] || untranslated[key] || key
+  args.forEach((arg, idx) => {
+    str = str.replace(new RegExp(`%${idx + 1}\\$s`, 'g'), String(arg))
+  })
   args.forEach(a => { str = str.replace('%s', String(a)) })
```
veloce: It would be nice to check whether this line can be removed. Since we pull new translation from crowdin, all string should have the new format.
If old format strings remain after the merge, I'd suggest to check if they are used in the app.
If we really must keep the old format, then let's keep this replace, but having a single `args.forEach`, not 2 (by tweaking the regex it is probably possible to have a since `str.replace` too).

jas14:
> Ok. I saw in the diff that this adds a lot of new strings to the JSON files.
> So the next step is probably to identify parts of the app where we can use those new strings (if they fit in the smaller space of the mobile app)

Sounds good! I'll chip away at that and would love some help.

veloce:
@jas14 thanks for this work! I'll merge it in the capacitor branch (new major version of the app) and start the work based on it (with some probable modifications). 
Capacitor branch already has heavy changes, so this is the good place to put the new translation system.

veloce:
Merge in capacitor branch: https://github.com/veloce/lichobile/pull/1055/commits/3592bf0d0b9ae5e4ce192433af8d08a6c4476906

Closing then.

PR #1058, opened 2019-10-01
Extend watch-scripts gulp task to rebundle for the platforms directories

ewanmellor:
Extend the watch-scripts gulp task to rebundle the version of app.js
(uglified) that goes in platforms/{android,ios}/www.

For example, this means that if a developer is using Xcode to run the
app, they can use "npm run watch-stage" and any changes in the source
code will get rebundled into platforms/ios/www/app.js automatically.
The developer still needs to restart the app inside Xcode (to copy
the files into the simulator and restart the app) but this is a faster
turnaround than "npm run prepare-ios-stage".

I have implemented this as a glob for platforms/{android,ios}/www.
I don't actually know if this is the correct workflow for android
developers using IDEs, but I presume that it is.  If not, we can
change the glob to just cover platforms/ios.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,3 +1,4 @@
+const glob = require('glob');
 const path = require('path');
 const source = require('vinyl-source-stream');
 const minimist = require('minimist');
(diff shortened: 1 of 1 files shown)


veloce:
I understand why it is useful, but I'm not sure I want to merge this actually. 
I know the current dev experience is not the best, but I'm in the process of migrating this app to [ionic capacitor](https://github.com/veloce/lichobile/pull/1055). 
This is a big change, so I don't know yet when it will be finished but it will fix a lot of issues regarding build. For a start the web app will always be served from a local server (CORS is now fully implemented on lichess server). Thus you can configure the capacitor app to directly load the assets from this local server during development.

ewanmellor:
I don't know anything about Capacitor, but it sounds goo.  I don't mind either way, you could merge this in the meantime, or just wait until the Capacitor work is done.

If you just want to close this, that's fine by me.

veloce:
I'll wait for the capacitor work indeed.

Issue #1051, opened 2019-10-01
Make Username In Inbox Conversation Tappable to Open Profile


landry314:
- Feature Request

Currently, while looking at a conversation with a user in the Inbox, there is no way to go to their profile to see what their stats are or to challenge them to a game. If they aren't being followed then it would be annoying to find them in the system.

Making it so that tapping a username goes to their profile would help a lot.

PR #1066, opened 2019-10-18
Make player names in message thread clickable

freefal:
I also discovered that the back button isn't shown if the player is the user himself, but this doesn't make sense to me. Why not still let the user navigate backward to where he came from even if he's looking up himself? I've removed that test and now the back button is always shown. I can revert this change if there's a reason for this behavior.

Closes https://github.com/veloce/lichobile/issues/1051

### DIFF
--- a/src/styl/inbox.styl
+++ b/src/styl/inbox.styl
@@ -95,8 +95,10 @@ table.threadList
       >.arrow
         vertical-align middle
         line-height 1
-      >.user
-        display inline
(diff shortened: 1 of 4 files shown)


jas14:
@freefal , thanks for the code. The messaging system has been overhauled since you wrote this PR - is the issue this fixes still present in the new version?

freefal:
New version looks great. Closing...

PR #1100, opened 2019-12-22
Some i18n cleanup

jas14:
Applied i18n to strings that are already translated in the lichess project, but were not in use here.


veloce:
Thanks.

PR #1109, opened 2020-01-17
made npm run build os agnositc/ runnable under windows, too

mkurz01:


veloce:
Sorry for the long delay on this. I've fixed it, slightly differently here https://github.com/veloce/lichobile/commit/c3cb6160934745099341988a0a5e6ac3e68787cf

PR #1115, opened 2020-01-27
lichess pieces neo

malferov:
lichess pieces neo by denguitarist
copied & converted from the source
https://userstyles.org/styles/135464/lichess-pieces-neo-modified


veloce:
Sorry I cannot accept it. This piece set is protected by a copyright.

PR #1177, opened 2020-04-07
Imperceptible lossy png optimization (pngquant)

niklasf:
Just did something similar for lila. Almost all images have a very reduced color palette, so the savings are quite significant (well ... at least relatively).


niklasf:
Also fixed a wierd black spot on the splash screen, reported here: https://lichess.org/forum/lichess-feedback/black-dot-on-mobile-android-lichess-app-at-startup

veloce:
Nice!

PR #1180, opened 2020-04-08
Change CSS to accomodate new 3x4 grid (vs. old 3x3 grid)

freefal:
Issue: The bottom row is shorter for the new 3x4 grid

old

Solution: Just specify that the grid has three columns, but don't specify the number of row. This gracefully handles both the new 3x4 grid

new (4 rows)

...and the old 3x3 grid.
new (3 rows)

My dev machine downloaded the new 3x4 grid and presumably people's phones will do the same as their caches expire (I think) so would be good to make this CSS change now so the UI can gracefully handle the switchover from 3x3 to 3x4.

### DIFF
--- a/src/styl/new-game.styl
+++ b/src/styl/new-game.styl
@@ -45,7 +45,6 @@
   flex-grow 1
   display: grid
   grid-template-columns: repeat(3, 1fr)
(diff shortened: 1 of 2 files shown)


veloce:
Thanks!

PR #1211, opened 2020-04-30
Added 'libra' piece set from lichess site.

greatwolf:
Another modern piece set, it doesn't look like it's active on lichess.org yet but it is in the master branch of lila's repo: https://github.com/ornicar/lila/tree/master/public/piece/libra.


veloce:
I won't merge it as long at it's not displayed in the website, sorry.

greatwolf:
Fair enough. This set can be revisited later when it does finally go live.

You should be able to merge the 'staunty' #1210 set though, that one is active and definitely live on lichess.org.

Issue #1212, opened 2020-05-01
Clock timing not playing soundfx for some modes.


greatwolf:
# Bug report

**Current behavior:**
The clock feature plays no sound when hitting the button and when flagging. This happens with most modes. The only modes that seem to work are Simply(Sudden Death) and Increment.

**Expected behavior:**
There should be sound when hitting the clock button or flagging for all modes. Or at least give the user an option to turn sounds on/off.

**Steps to reproduce:**
- Open lichess app
- Tap the upper left corner to open menu
- Scroll down to 'Tools' and tap 'Clock'
- Set the timing mode to anything other than 'Sudden Death' and 'Increment'
- Start the clock and tap the buttons. There's no sound.

**Environment information**

- phone hardware: OnePlus One
- phone operating system version: Android 6
- version of lichess mobile: Lichess 7.0.4

**Other information:**

It looks like sound was turned off for all the other modes in this commit:  [`5371a7e8a72764f0c40296251d3481547f53ff8e`](https://github.com/veloce/lichobile/commit/5371a7e8a72764f0c40296251d3481547f53ff8e#diff-be5e98d1f4a84d2acb40c58e2f405476R29)

There's no explanation offered as to why other than the commit message: "Clock simplification and fixes"

greatwolf:
I've submitted PR that reenables this. Hopefully this gets merged in without too much controversy.

PR #1213, opened 2020-05-01
Reenabled sound effects for other clock modes. Fix for issue#1212

greatwolf:
This should fix issue #1212 where sound effects aren't playing for many modes.


veloce:
Thanks!

Issue #1231, opened 2020-05-12
To add a new theme from lichess website


LichessUser:
![IMG_20200512_091118](https://user-images.githubusercontent.com/64771561/81638047-8a23c900-9435-11ea-9169-c397dd10b350.jpg)

**I'm submitting a ...**  (keep the relevant one):
- bug report
- feature request

# Bug report

**Current behavior:**

**Expected behavior:**

**Steps to reproduce:**

**Environment information**

- phone hardware
- phone operating system version
- version of lichess mobile

**Other information:**

# Feature request

This theme is really cool, I want this theme on the lichess app as it does not has any good pieces and board sets, so please add this theme to the app.

greatwolf:
Is this referring to the board or the piece set or both? What's the name of them?

LichessUser:
How to see the names of the pieces?

greatwolf:
click on piece set or board themes and just mouse over it should show the name.

LichessUser:
Any way to show it on mobile, I don't have a pc

LichessUser:
> click on piece set or board themes and just mouse over it should show the name.

These are the board and pieces I am talking about
![IMG_20200513_021945](https://user-images.githubusercontent.com/64771561/81744106-648fd180-94c0-11ea-87e5-06415907fa12.jpg)
![IMG_20200513_021916](https://user-images.githubusercontent.com/64771561/81744149-740f1a80-94c0-11ea-98b9-01d5dbd11831.jpg)

LichessUser:
> Is this referring to the board or the piece set or both? What's the name of them?

Board name:- Wood4
Pieces name:- Cardinal

Please add these to the lichess app.
Thank you

LichessUser:
Or you can tell me way to add these board and pieces myself

LichessUser:
Can you do anything for me Me GreatWolf regarding this?

veloce:
Hello, I'm planning to add all board themes in the next release.

veloce:
For missing piece theme, I'll accept PRs.

LichessUser:
Sir when will the next update come?, And what is meant by pRS?

LichessUser:
> For missing piece theme, I'll accept PRs.
Please add all the pieces too, which are available on website.
It would be really helpful.
Thank you

LichessUser:
> For missing piece theme, I'll accept PRs.

Please add all the pieces too, which are available on website.
It would be really helpful.
Thank you

[...]

Issue #1230, opened 2020-05-11
Feature request: Add all board themes + piece sets that are on lichess website


Ejstn:
I propose to add all board themes + chess pieces that are on the lichess website but are missing in mobile application.

Namely: there are **24 board themes on web but 12 on mobile**, there are **22 piece sets on web and 16 on mobile**.

My argument / use case is I would like to set the same theme + pieces on web and on mobile, so if I change between web and mobile it is more convenient / seamless.

LichessUser:
I agree, they should add all the theme, I am bored with the themes on app

LichessUser:
I l
![IMG_20200512_091118](https://user-images.githubusercontent.com/64771561/81640643-af680580-943c-11ea-88a8-529e5b5f5cd4.jpg)
ove this one

PR #1237, opened 2020-05-14
Add extra missing lichess board themes.

greatwolf:
This PR adds the missing lichess.org board themes to the mobile app. This goes towards addressing #1230 and #1231 as requested by other users.


Ejstn:
Awesome, thanks!

veloce:
Many thanks for this @greatwolf . Though, I am sorry, as I won't be able to merge your PR as is. The thing is that I don't want to add that many boards to the app package, as I try to keep it light. I have already a basis to download themes and save them to the app and I have some local work (not pushed yet) already done to do the same for boards.
But what you did here will be useful nevertheless, especially the css file, so I'll try to integrate your commit in my branch.

greatwolf:
sure thing.

veloce:
One PR per piece set is fine yes. Pieces are svg so it's much lighter, so I don't plan to remove pieces from the package for now.

LichessUser:
So you mean you will add new piece sets and won't remove any elder one as they are really lite?

PR #1246, opened 2020-05-18
Added more badges.

greatwolf:
Because you can never have too many badges


LichessUser:
Hey GreatWolf, are you a app developer?
And which badges are you talking about?

PR #1247, opened 2020-05-19
Added 'maestro' piece set from lichess site.

greatwolf:


LichessUser:
Hey GreatWolf, I am having some issues with the cardinal pieces, are you having the same?
Veloce told me that my device is a low end device that's why it's lagging, do you have any solution for it?

greatwolf:
It looks like the slowdown's coming from filter use in the svg. Removing the filter will remove the drop-shadow of the pieces but it should fix the performance issue.

LichessUser:
Can you please add a PR for it?

Issue #1214, opened 2020-05-01
Over the board handicap increment results in flags prior to end of time


aestrivex:
**I'm submitting a ...**  (keep the relevant one):
- bug report

# Bug report

**Observed behavior**
In over the board chess, with the handicapped increment time control, my wife and I played several games where white had 6-3, black had 3-2. The games ended at apparently random times where white lost due to being out of time, despite having minutes left on the clock.

"**Expected behavior**
The expected behavior is, no player loses on time unless the time has expired.

**Steps to reproduce**
Play over the board with handicap increment 6-3 and 3-2. Three games at this time control flagged on time. White flagged each game with at least 2 minutes left on the clock. I tried playing test games with stupid moves at 2-1 and 1-1 and was unable to reproduce. 

**Environment information**

- phone hardware
- phone operating system version
- version of lichess mobile

Playing on Alcatel Idol 5S [6060S]
Android 7.1.1 build v4E5P, Linux kernel 3.18.31 

**Other information:**

greatwolf:
After looking into this issue a bit, I'm able to reproduce the bug with a high degree of probability.

 - Set the clock to 30 seconds each, 0 increment.
 - Start the game and let the clock run for a couple seconds.
 - Now tap the circle + to start a new game.
 - Set the time to something higher than the last. eg. 1 min.
 - Start the game again and let the clock run.

It will eventually timeout even though there's still time left on the clock. Handicap is incidental and does not appear to be directly relevant. This bug can happen whether handicap is used or not.

The cause of the issue appears to be the clock event continuing to tick even though the new game menu is open. When a new game starts the old clock from the previous game still continues to fire and that causes the erroneous flagging.

Will be submitting a PR shortly to fix this.

PR #1250, opened 2020-05-21
Fix for premature flagging for OTB clock.

greatwolf:
This PR should fix the issue reported in #1214. It does so by stopping the OTB clock from ticking when the user taps and opens the 'newGameMenu'.


veloce:
Thanks!

PR #1259, opened 2020-05-25
Fix a typo in Hebrew localisation

homoSapiensSapiens:
The word for moves played is מהלכים, not מהלים.


veloce:
Sorry, this is a generated file that should not be modified manually. If you want to contribute to translate lichess, this is where it is happening: https://crowdin.com/project/lichess

PR #1263, opened 2020-05-29
Add colors to tournament list page

freefal:


src/lichess/interfaces/tournament.ts
```diff
@@ -158,11 +158,13 @@ export interface TournamentLists {
 }
 
 export interface TournamentListItem {
+  readonly battle?: {}
```
veloce, line 161: why empty object as type? the best would be to type it, or you can use `any` for now.

freefal:
When I looked at the tournament list the non-team battle tournaments were
missing the `battle` field and the team battle tournament had a `battle`
field, but it was just any empty object. Thus the type signature you see
here. We're just testing for the presence/absence of this field so its type
doesn't matter.

Let me know what you'd like me to do. I'd be happy to change to `any` if
you think that's better.

veloce:
Ok let's keep the empty object type.

freefal:
FYI, I found some infrequently used tournament colors here (https://github.com/ornicar/lila/blob/1b66b22f7b61f9f6eae37fe76de6cd8c958b1715/ui/tournamentSchedule/css/_schedule.scss) and added them.

Otherwise, should be done pending resolution of the type issue on the battle field.

freefal:
Sounds good. I pinged thibault too. Could just be an error on the server
side and that this was supposed to be a boolean. Can change later if he
confirms.

PR #1266, opened 2020-05-31
Fix error with plurals

asivers:
Hello! I have found a small bug in lichess app playing in Russian: plurals were not displayed correctly. Only "one" and "other" plurals were recognized, though in Russian we have also "few" and "many" plurals. The mistake was in quantity function in src/i18n.ts. In any case this function returned default value. I changed names of locales in switch statement ("ru" -> "ru-RU" and for other languages) to fix this problem. Testing showed that now plurals are displayed correctly.

### DIFF
--- a/src/i18n.ts
+++ b/src/i18n.ts
@@ -198,57 +198,57 @@ function quantity(l: string, c: number): Quantity {
   const rem10 = c % 10
   switch (l) {
     // french
-    case 'fr':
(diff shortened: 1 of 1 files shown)


veloce:
Thanks for catching this. I've fixed it using another way: https://github.com/veloce/lichobile/commit/f625ea4f53481e71ce0f8a2de032ab05f1fc12ab

Issue #1074, opened 2019-10-24
Bug: "Playing streak" always shows last played date


KasperDeBlieck:
# Bug report

**Current behavior:**

The field "playing streak" shows the date of the last played game instead of the date were the playing field happend (/src/ui/user/perfStats/variantPerfView.tsx)

**Expected behavior:**

Field showing the right date.

**Steps to reproduce:**

**Environment information**

- Iphone X
- iOS 13.1.3
- 6.3

Tay0108:
Same on Android, app version: v7.1.2. On web it shows correct date.
![101405420_295645968498855_2876627541328658432_n](https://user-images.githubusercontent.com/10603901/83402944-0479b480-a408-11ea-901b-3943c1e4dc26.png)
<img width="1105" alt="Screenshot 2020-06-01 at 12 55 45" src="https://user-images.githubusercontent.com/10603901/83402947-06dc0e80-a408-11ea-9eea-7f4957f5c2d0.png">

PR #1271, opened 2020-06-02
fix(user profile): playing streak date

Tay0108:
Ref. https://github.com/veloce/lichobile/issues/1074

I've changed `variantPrefView.tsx` to show playing streak from date instead of last game played date. See referenced issue for more details.


veloce:
thanks!

PR #1309, opened 2020-07-13
Escape new lines and add forum link

freefal:


src/ui/menu/menuView.tsx
```diff
@@ -193,6 +193,11 @@ function renderLinks(user?: Session) {
         <span className="fa fa-at"/>{i18n('players')}
       </li> : null
       }
+      {online ?
```
veloce, line 196: I think it's still not ready to be released. So let's remove it from the menu for now. I should add a project with a todo list of what needs to be done before release, asap.

freefal: Ok, just removed.

PR #1324, opened 2020-07-27
Set unzip maxBuffer to 10MB 

freefal:
Set maxBuffer to 10MB when executing unzip to avoid errors encountered when default 1MB used

Found a similar issue and fix here: https://github.com/nicojs/node-install-local/issues/8


eoyx12 (commented):
ال

Issue #374, opened 2016-07-07
Conditional premoves in corres games


veloce:

lightbulbjim:
+1

bak1an:
I am interesting in having this, will try to create a PR implementing that.

c5gaspar:
> I am interesting in having this, will try to create a PR implementing that.

Any updates on this PR?

bak1an:
@c5gaspar I started working on it back then but had to abandon it due to a lack of free time.

Thanks for pinging, it is a good reminder to dedicate some hours to finishing half-baked PRs.

KennethJ-USC:
Still really interested in this feature. What would you estimate the amount of dev time it would take to finish your implementation @bak1an ?

PR #1464, opened 2020-12-24
Convert tests to Typescript

jas14:
I'm working on #374 and am writing some unit tests. I can't get the current custom Jest preprocessor to work with TypeScript declaration files, so, figured I'd type the existing tests and use `ts-jest`.


.travis.yml
```diff
@@ -11,7 +11,7 @@ android:
       - extra-android-m2repository
 sudo: false
 before_install:
-  - nvm install 10.16.3
+  - nvm install 12.2.0
```
jas14, line 14: Node 10 for some reason errored on `import 'whatwg-fetch'`, but Node 12 didn't. The README says the Node dependency is "latest LTS," which I guess is 14.x, but I'm hoping the latest 12.x is OK here.

tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "noUnusedLocals": true,
     "jsx": "react",
     "jsxFactory": "h",
-    "types": ["lodash-es", "d3"]
+    "types": ["lodash-es", "d3", "jest"]
```
veloce, line 18: I'd like not to add jest types to the main tsconfig.json. It looks like it's possible to define a custom tsconfig for jest (that extends the global one) by modifying jest config like so:
```
{
  ...
  "globals": {
    "ts-jest": {"tsConfigFile": "tsconfig.test.json"}
  }
}
```

veloce (changes requested):
Thanks! looks fine except for the tsconfig modif.

src/dts/index.d.ts
```diff
@@ -6,3 +6,6 @@
 /// <reference path="signals.d.ts" />
 /// <reference path="misc.d.ts" />
 /// <reference path="mithril.d.ts" />
+
```
veloce, line 9: node types are not used so this should be removed

veloce (changes requested):
just one more change, and I'll merge it

src/dts/index.d.ts
jas14: Whoops. fixed

PR #1475, opened 2021-01-01
RFC: Migrate from TSLint to ESLint

### LINKED ISSUE #1467: Migrate to ESLint from TSLint?
Palantir, the maintainer of TSLint, [announced](https://medium.com/palantir/tslint-in-2019-1a144c2317a9) in Feb 2019 that TSLint is deprecated in favor of ESLint. The latter will have better continued support, so it may be worth switching.

jas14:
Closes #1467.

The linters can be configured to have different warnings levels, enforced styles, or can be disabled completely. If you'd like to see what linters raised errors initially, check out 
cf26093 and run `npx eslint src/ --quiet`.

Lastly, there are still tons of warnings, which don't currently cause linting to fail but will appear in `npm run lint` output and should be incrementally addressed in the future.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,5 +14,6 @@ module.exports = {
   },
   "plugins": ["@typescript-eslint"],
   "rules": {
+    "@typescript-eslint/triple-slash-reference": "off"
   }
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,18 @@
+module.exports = {
+  "env": {
+    "browser": true,
+    "es2021": true
+  },
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended"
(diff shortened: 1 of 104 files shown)


veloce:
Whaa! Many thanks for this!! And thanks for the added test. I think I'll merge it as is. The only changes that I fear a little are the regex escape changes since the regex were working well with the escape. So I'm gonna test that first then merge it.

veloce:
Oh, before merging the travis build needs to be fixed. Looks like the new backoff test doesn't pass here.

jas14:
Just fixed locally and should pass CI. As for the ["unnecessary escaping"](https://eslint.org/docs/rules/no-useless-escape) fixes in regexes, I can downgrade that error to a warning - just let me know.

veloce:
Many thanks again @jas14

PR #1477, opened 2021-01-02
Add california piece set

### LINKED ISSUE #1026: California piece set on the site, but not in the app.
**I'm submitting a ...**  (keep the relevant one):
**- feature request**
# Feature request

Add the California piece set to app please so that everything can look the same between devices <3

isaacphi:
Closes #1026 

:wave: Tiny contribution here. The "California" piece set is available on web but not on the app so here it is

california-piece-set

I more or less just followed this commit as an example https://github.com/veloce/lichobile/commit/51fca5a46b590f0776fdb9ab7fe25ecfa9cd9935

It looks like the `gioco` and `icpieces` piece sets are also missing. I'm more than happy to add those in too either in this PR or another one. I just wanted to confirm that this looks good first.

This is my first contribution so lmk if I missed anything! One thing I was unsure of is the ordering of the piece sets.
:chess_pawn::chess_pawn::chess_pawn::chess_pawn::chess_pawn::chess_pawn::chess_pawn::chess_pawn:


veloce:
Thanks :)

PR #1490, opened 2021-01-12
Conditional Premoves

### LINKED ISSUE #374: Conditional premoves in corres games

jas14:
I'm pleased to present an implementation of conditional premoves on mobile! Closes #374.

https://user-images.githubusercontent.com/569991/104317254-b6ec1c80-54ab-11eb-99c8-ea0977ff815a.mov

https://user-images.githubusercontent.com/569991/104317252-b5baef80-54ab-11eb-9848-ac01b795bdbc.mov

Open to any and all feedback, especially design-related.


veloce:
Wow! looks awesome, thank you. I'll take the time to test it, I'll get back to you asap.

veloce:
@jas14 I have a first feedback based on the videos. 

I prefer not to have another tab for this feature, because it's not the most convenient here, and I don't like the eye icon. 
I think we should use the same kind of overlay window (dunno how to name this properly...) that we have for the "practice vs. computer" or "Learn from your mistakes" features.
The difference is that the window cannot be closed (if the corres. game is still playable of course, otherwise you got the regular analysis), but you can only toggle between expanded/minimised. 
When you click "analyse" from a corres. game we need to decide if the conditional premoves window is already expanded, or not. I'm not sure which is the best yet, I guess expanded is fine.

jas14:
@veloce sounds good. like this?

veloce:
@jas14 looks great! yes this is exactly what I had in mind: so when you go to analysis from corres. game, conditional premoves are directly accessible (and visible).
I'd probably change the analysis button location for corres games too (in the tabbar instead of in the menu).

jas14:
All right! I decided to start with conditional premoves collapsed. You'll notice that when it's expanded, it hides the whole analysis table. I think existing users would find it jarring to find their favorite analysis features hidden by the overlay if it were expanded to begin with. If you strongly disagree, I can make it expanded by default with just a one-line change.

I forgot to do theme styling in the last round too, so here's a video with the light theme. The dark theme looks pretty much the same as it used to.

https://user-images.githubusercontent.com/569991/104822217-e766ea00-580e-11eb-9604-357aa8639988.mov

https://user-images.githubusercontent.com/569991/104822215-e6ce5380-580e-11eb-91d3-6ce613a916f1.mov

veloce:
Looks very nice! I'm ok with collapsed, the popup bar is visible so I guess it's fine.
I have 2 more suggestions to improve the UX:
- as in the website I'd put the "play" icon at the left of a recorded variation so you know that it's been save, and it's going to be played
- also as in the website, I'd put the "red cross icon" at the very right, so when you tap this icon it it will delete the variation (without showing the confirmation popup)
- you can still keep the long press to show the popup to delete the variation

veloce:
And, another one, the icing on the cake :)

Like in the website, it'd be nice to know if you recorded variations when you are on the game screen. For that we need to put the analysis button in the bottom tab bar. I'd say put it in place of the notes (or switch side?) button that will be put in the menu instead. Then we put over analysis button the number of recorded variations in the same kind of chip that is used for unread chat messages. I suppose that information is available somewhere in the game json, since it's used by the website.

jas14:
Sounds good. I especially like the analysis icon w/chip on the round view tab bar!

> * also as in the website, I'd put the "red cross icon" at the very right, so when you tap this icon it it will delete the variation (without showing the confirmation popup)
> * you can still keep the long press to show the popup to delete the variation

I'm having trouble getting these to work together, so I think I will abandon the long-press contextual menu. I'm also planning on having the red cross icon appear only when the forecast is tapped - otherwise, I think it would be too easy to  accidentally press the wrong one

veloce:
> I'm having trouble getting these to work together, so I think I will abandon the long-press contextual menu. I'm also planning on having the red cross icon appear only when the forecast is tapped - otherwise, I think it would be too easy to accidentally press the wrong one

yeah I think it's even better like that

src/ui/analyse/forecast/forecastView.ts
```diff
@@ -0,0 +1,164 @@
...
+          'div.forecasts-list',
+          {
+            class: settings.game.pieceNotation() ? 'displayPieces' : '',
+          },
+          [
+            ...fctrl.lines.map((nodes, i) => {
+              return h(
+                'div.forecast[data-icon=G]',
+                {
+                  key: `${i}-${nodes.map(node => node.san).join('')}`,
```
jas14, line 103: I know, this basically defeats the purpose of keys... not proud of it. The issue is that if we just use the SAN, the ontap handler may have the wrong index if a forecast is deleted.

Ideally `ForecastCtrl` would maintain a map from SAN to `ForecastStep[]`, which would keep the key here constant and would let Mithril do its smart diffing magic.

jas14:
https://user-images.githubusercontent.com/569991/104982275-517ebb00-59d8-11eb-9600-6ae2b9ff9f76.mov

src/ui/analyse/forecast/forecastView.ts
jas14: fixed!

src/ui/analyse/forecast/ForecastCtrl.ts
```diff
@@ -0,0 +1,195 @@
+import { keyBy, pickBy, some, values } from 'lodash-es'
```
veloce, line 1: I try to keep my utilisation of lodash as low as possible, and when I use a lodash function, I'd import it like to `import stuff from 'lodash-es/stuff'` to avoid importing everything (tree shaking is not working here I believe).

But here I'd not use any of them (see comments below).

src/ui/analyse/forecast/ForecastCtrl.ts
```diff
@@ -0,0 +1,195 @@
...
+import { AnalyseDataForForecast, ForecastStep, MinimalForecastStep } from '~/lichess/interfaces/forecast'
+import router from '~/router'
+import redraw from '~/utils/redraw'
+import { playAndSaveForecasts, saveForecasts } from './xhr'
+
+const MAX_FORECAST_PLIES = 30
+
+type SanMap = {[san: string]: ForecastStep[]}
```
veloce, line 9: I'd use a `Map` here instead. This is a better data structure for this usage, and I think we can easily avoid all the lodash functions above

src/ui/analyse/forecast/ForecastCtrl.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * @returns whether the given forecast nodes are part of a new candidate premove line that is not
+   *          a prefix of the other saved conditional premove lines.
+   */
+  isCandidate(nodes: MinimalForecastStep[]): boolean {
+    const forecastCandidate = this.truncate(nodes)
+    if (!this.isLongEnough(forecastCandidate)) return false
+
+    const isPrefix = some(this._lines, (existingForecast) => {
```
veloce, line 59: easily reimplemented using `Array.prototype.some` and `Map.prototype.values`

veloce (commented):
Really awesome work! I have only a couple of changes to suggest.

jas14:
Thanks for the continued reviews!

veloce:
Sorry there will be a little delay to merge it as I need to focus on the current release with stockfish first. But I won't forget it of course :)

veloce:
Just added a couple of fixes (was missing a web socket event to listen to new moves) and UI tweaks on top of it. Thanks again.

Issue #1444, opened 2020-12-16
Some piece sets differ from web version


smiglo:
**Is your feature request related to a problem? Please describe.**
Some piece sets were updated in web version just after those sets were committed here:
* cardinal
* governor
* maestro

Pieces were committed here on Sep, 13 and updated in lila on Sep, 27.

**Describe the solution you'd like**
Could you update them to be in-sync between web and mobile version? :)

**Additional context**
The update was a result of this ticket in lila: https://github.com/ornicar/lila/issues/7167

PR #1507, opened 2021-01-17
Update piece sets to match web versions

janhennings:
This updates the Maestro and Cardinal piece sets to be in sync with the newest web versions.
Tiny contribution that addresses #1444


veloce:
Thanks!

PR #1520, opened 2021-01-25
Websocket chat

### LINKED ISSUE #1491: Switch from "compat" messaging to WebSocket-based messaging
**Is your feature request related to a problem? Please describe.**

The current mobile messaging system feels more like email than chat. lila has already implemented WebSocket-powered chat server-side, and uses it on the desktop site. Notable differences:
- Plain HTTP requests (reload required) vs WebSocket (real-time chat)
- Subject line required in current implementation for new messages

**Describe the solution you'd like**

For better parity with the desktop site and for improved ease of use, I'd like the mobile app to be able to do real-time messaging without a subject line.

**Describe alternatives you've considered**

Minimally, I'd like to add a placeholder for the body of the new message. There's a placeholder for the subject line but not the body, and it's difficult to see the separator between subject and body, which is confusing and has lead me and friends to fill out the entire message in the subject line. See screenshot.

As a short-term step to at least remove the subject line requirement, we might be able to use the official [Lichess messaging API](https://lichess.org/api#operation/inboxUsername), as the screen for replying to an existing thread already does.

**Additional context**

Here's the lila desktop version, which is what I'd like to more or less see on mobile (with different styles, probably):

jas14:
As promised, I have ~shamelessly ripped off~ _ported_ much of the messaging code from lila. Style is informed by the lila messaging module and, to some extent, the existing in-game chat. Closes #1491.

This PR aims to contain all of the features of the current non-game messaging subsystem, but does not have full feature parity with desktop; notably, I omitted the following:
- Text enhancements (aside from hyperlinks), e.g. images, Lichess game embeds
- The "report" action
- The "block" action
- The "challenge to a game" action

I imagine the last 3 would be fairly easy for anyone who felt like picking them up.

## Tablet demo
featuring main menu inbox count, unread messages, online/patron/title indicators, blocked/nonpostable UI, and typing indicators

https://user-images.githubusercontent.com/569991/105650924-60a3b400-5e83-11eb-83b0-2f30b2eea07b.mov

## Mobile demo
featuring new thread composition and autoscrolling

https://user-images.githubusercontent.com/569991/105650923-600b1d80-5e83-11eb-9cc1-f1b7ee4fc0d7.mov

## Mobile demo 2
featuring links and thread deletion

https://user-images.githubusercontent.com/569991/105650922-5f728700-5e83-11eb-823a-e82ea64722af.mov


veloce:
Very nice! Thanks a lot. Judging from the demos everything is there, I guess this will land in next release along with conditional premoves.

veloce (commented):
I started to test this and I have just a couple of general code style remarks for now:

- semicolons at the end of statements should be removed because I don't use them in the rest of the code (we should add that rule to the linter)
- I'd move all the "view" related files to a `view` folder, again to follow conventions in the rest of the app
- I don't use `class` property name, rather `className`; even though mithril supports `class` I'd prefer to use `className` to follow this code style convention like in the rest of the code
- don't know for sure it's relevant here, but we need to check that `ontap` button handler is used only in static elements. When you are in a element that scrolls, you should use `ontapY` instead.

When I tested the code I noticed that the scroller was not working all the time. And even at some point not working at all (was the case on ios simulator). The `oncreate` handler in `renderMsgs` was simply not executed, it seems. I'll try to debug this further if I have the time if you can't reproduce.

jas14:
Got it, thanks! I'll take care of the style points. As for scrolling, I have some repros, I'll push fixes in a bit

PR #1529, opened 2021-01-30
Use withoutNNUE flavor by default

cdietze:
Make the Android "withoutNNUE" flavor the default which makes building simpler and is probably what newcomers to the project want.
I tried it out, and after this fix "withoutNNUDebug" was selected when I created a new Android Studio project:
Screenshot 2021-01-30 at 21 38 36


veloce:
Thanks!

Issue #758, opened 2018-01-13
Add support for coordinate training in mobile app


bak1an:
If the feature is acceptable I could work on a PR to implement it.

veloce:
Sure! You can go ahead and implement it. I wasn't planning to do it by lack of time, but this is definitely acceptable.

aldokkani:
@veloce I am working on this and almost done, but I am having issue with "mithril" lib to have the component redraw/rerender when the user clicks. Can I get your help?

freefal:
Hossam,

@veloce is better qualified to help here, but having run into mysterious
mithril issues in the past, make sure that you are adding unique "key"
attributes to any clickable HTML element.

aldokkani:
@freefal I am not sure if this is the problem, since I am using the board and the board should already have a key, shouldn't it?

veloce:
Maybe you miss a `redraw()` call? I know this is something now obvious in mithril (or it's also the way I use mithril), but except for the button handlers (`helper/ontap`) each time an action is performed and you want to redraw, you need an explicit call to `redraw()`.

PR #1544, opened 2021-02-08
feat: coordinates training #758

aldokkani:
create coord logic and view

move logic to a controller

fix board orientation,

Update translation

change back wrong answer state


aldokkani:
#758

www/i18n/en-US.js
```diff
@@ -348,6 +348,9 @@ export default {
   "automaticallyProceedToNextGameAfterMoving": "Automatically proceed to next game after moving",
   "autoSwitch": "Auto switch",
   "puzzles": "Puzzles",
+  "coordinates": "Coordinates",
+  "coordinates-training": "Coordinates Training",
+  "startTraining": "Start Training",
```
jas14, line 353: Our translations are sourced from [CrowdIn](https://crowdin.com/translate/lichess), so just adding these translations in this locale file won't do anything. Fortunately, these keys are already translated :)

You'll need to add the `coordinates` module here: https://github.com/veloce/lichobile/blob/b7c3a0b8f1d304f3ee5cdb639c6a6bee82ea74be/scripts/i18n-update.ts#L12

and then re-run the script, but then you'll get these translations for free. (Note that the keys may be different.)

src/ui/menu/menuView.tsx
```diff
@@ -174,6 +174,11 @@ function renderLinks(user?: Session) {
           <span data-icon="4" />{i18n('studyMenu')}
         </li> : null
       }
+      {online ?
+        <li className="side_link" data-route="/coord">
+          <span data-icon="4" />{i18n('coordinates')}
+        </li> : null
+      }
```
jas14, line 181: Does the app need to be online for coordinates training?

www/i18n/en-US.js
veloce: 👍

src/ui/menu/menuView.tsx
veloce: This feature should be available offline yes. It's just that when offline we don't send the score to the server.

aldokkani: @jas14 no it doesn't. It was a mistake, I forgot to remove it after fiddling with the code at first.

www/i18n/en-US.js
aldokkani: Thanks, fixed!
But could you elaborate more how does this work? How by adding the module name to that array I got the correct translation?

jas14: Sure, the code in `scripts/i18n-update.ts` basically does the following:
1. Download and unzip the complete `lila` translation package, which is organized into "modules" (which are just separate translation files + folders)
2. Copy translations from only the modules (folders) that we use, for performance/optimization purposes

So adding the `coordinates` string to that array indicated that the script should also import the translations from the `coordinates` module/folder. That's all :)

aldokkani: I see, this is lila translation which is used by lichess web as well, also `coordinates` is the name they use in this package not the name of the module I wrote.

aldokkani:
It is worth mentioning I was imitating the web feature as much as I can, one thing I couldn't imitate is the **transition animation effect** when a user answer correctly moving to the next coordinate.

jas14:
@aldokkani you may need to merge `master` into this branch (or rebase onto `master`) in order to pick up the [new eslint rule](https://github.com/veloce/lichobile/pull/1542). `npx eslint --fix src/` will fix the new lint errors.

aldokkani:
I run npm lint before creating the PR, u saying this in general or you saw
lint error?

jas14:
Since https://github.com/veloce/lichobile/pull/1542 was merged, there are 66 new lint errors.

veloce:
Sorry if I'm a bit long to review this PR, as I need to take care of the websocket chat one first, but I won't forget it don't worry!

veloce:
@aldokkani before I start to review the PR, I'd need you to rebase id on master to get the latest update.
I added a new eslint rule to enforce single quotes. I saw that your PR uses a lot of double quotes and I would like the project to be consistent with that.
So if you rebase on master you'll be able to run `npm run lint -- --fix` to automatically turn them to single quotes.
Thanks!

aldokkani:
@veloce done

src/ui/coordinates/coordView.ts
```diff
@@ -0,0 +1,103 @@
...
+import Board from '../shared/Board'
+import i18n from '../../i18n'
+import layout from '../layout'
+import { INITIAL_FEN } from 'chessops/fen'
+import CoordCtrl from './coordCtrl'
+
+let orientation = (Math.random() > 0.5 ? 'white' : 'black') as Color
+let random = true
+
+function reOriente(color: Color | undefined): void {
```
veloce, line 14: that function could be in the controller I think

src/ui/coordinates/coordView.ts
```diff
@@ -0,0 +1,103 @@
...
+      h('div.coord-trainer__side', { style: { visibility: ctrl.started } }, [
+        h('div.box', [h('h1', i18n('coordinates'))]),
+        h('form.color.buttons', [
+          h('group.radio', [
+            h('div', [
+              h('input#coord_color_3', {
+                type: 'radio',
+                name: 'color',
+                value: 3,
+                onclick: () => reOriente('black'),
```
veloce, line 57: We shouldn't use `onClick` at all in the app to be consistent with the other buttons. Instead use `ontap` handler, from `~/ui/helper/index.ts`.
Ex usage here: https://github.com/veloce/lichobile/blob/master/src/ui/home/homeView.tsx#L211

src/ui/coordinates/coordView.ts
```diff
@@ -0,0 +1,103 @@
...
+      h('div.coord-trainer__board.main-board', [
+        ...ctrl.coords.map((e: Key, i: number) =>
+          h(`div#next_coord${i}.next_coord${isWrongAnswer(i)}`, { key: i }, e)
+        ),
+        board,
+      ]),
+      h('div.coord-trainer__table', { style: { visibility: ctrl.started } }, [
+        h(
+          'button.start.button.button-fat',
+          { onclick: ctrl.startTraining.bind(ctrl) },
```
veloce, line 93: I'd rather not use `bind` because this is not type safe. Use arrow function instead.

src/ui/coordinates/coordView.ts
```diff
@@ -0,0 +1,103 @@
...
+                onclick: () => reOriente('white'),
+              }),
+              h('label.color.color_1', { for: 'coord_color_1' }, [h('i')]),
+            ]),
+          ]),
+        ]),
+      ]),
+      h('div.coord-trainer__board.main-board', [
+        ...ctrl.coords.map((e: Key, i: number) =>
+          h(`div#next_coord${i}.next_coord${isWrongAnswer(i)}`, { key: i }, e)
```
veloce, line 86: I usually don't use complex string interpolation in mithril hyperscript function 1st param. I believe this is not the most efficient. Usually you'd use a static string, and put the rest in the object, like this:
```
h('div.next_coord', {
  key: i,
  id: 'next_coord' + i,
  className: ctrl.wrongAnswer && i === 0 ? '.nope' : '',
}, e)
```

If you have several classes to set, you can use `classSet` from `~/ui/helper/index.ts`.

veloce (changes requested):
A couple of things that I noticed should be fixed.

src/ui/coordinates/coordView.ts
aldokkani: Yea this better and simpler, will update it.

src/ui/coordinates/coordView.ts
aldokkani: It modifies the view property `orientation`  and then redraw, I don't think it can be moved without a lot of complications. But more importantly I don't think it belongs to the controller.

aldokkani: I moved the code to chessground and used similar functions

src/ui/coordinates/coordView.ts
aldokkani: `ontap` works only with buttons these are inputs radio, the problem is when I changed them to buttons when I tap them they navigate out the home page I couldn't debug the problem

src/chessground/Chessground.ts
```diff
@@ -161,6 +161,10 @@ export default class Chessground {
     anim(board.toggleOrientation, this)
   }
 
+  orienteWithColor = (color: Color): void => {
```
veloce, line 164: I'd prefer not to have that change in chessground, it was fine the way you did it previously

src/chessground/util.ts
```diff
@@ -78,6 +78,10 @@ export function opposite(color: Color) {
   return color === 'white' ? 'black' : 'white'
 }
 
+export function randomColor(): Color {
```
veloce, line 81: no need to put that in chessground, as only the coord view will use it.

src/ui/coordinates/coordView.ts
```diff
@@ -74,7 +42,7 @@ export default function view(ctrl: CoordCtrl) {
                 type: 'radio',
                 name: 'color',
                 value: 1,
-                onclick: () => reOriente('white'),
+                onclick: () => ctrl.chessground.orienteWithColor('white'),
```
veloce, line 45: you can use `renderRadio` or `renderMultipleChoiceButton` from `src/ui/shared/form.ts`  instead of this.

veloce (commented):
please revert your recent changes in chessground, and see my comment to see how to fix the `onClick` issue.
thanks

src/ui/coordinates/coordView.ts
aldokkani: I used renderRadio but I have to make some changes to it.

src/chessground/Chessground.ts
aldokkani: I do not know why would you prefer the code not to be centralized and modular depending on the functionality, meaning handling all the Chessground logic in the Chessground module.
But I do prefer it that way, I think I was doing it wrong previously.

PR #1549, opened 2021-02-10
Add horsey piece set

Reepeecheep:


veloce:
Thanks, and I added the board as well!

Issue #1539, opened 2021-02-06
Lichess will restart in N minutes


ddugovic:
**Is your feature request related to a problem? Please describe.**
Players lose their games after a server restart:
> I have never seen an orange "server will restart" message on my android app screen - did I blink and miss it?? In any case it should not be upto the user to observe and reset communication glitches that the app developers are aware can happen from time to time.

**Describe the solution you'd like**
Display when a server restart is about to occur; even if the application cannot handle the restart, the player can troubleshoot a connection problem.

**Describe alternatives you've considered**
Make the app resilient to Lichess server restarts somehow.  I don't know.

**Additional context**
> I am using the ap on my android tablet and I am waiting for my opponent to move. Looks like they lost on time - there are no indication of connection problems on my screen - both me and my opponent have a green dot- of course when I check if a result has been reached I see I lost. If I need to refresh my screen how do I do it? Why should I need to refresh without a warning? This issue is really annoying as it has happened to me multiple times and always when I might be winning and really frustrating in tournaments.

https://lichess.org/forum/lichess-feedback/lost-on-time-while-waiting-for-them-to-move-again

PR #1563, opened 2021-02-20
Show announcements

jas14:
Part of https://github.com/veloce/lichobile/issues/1539 .

This does not include:
- real-time `announce` message handling on websockets
- announcements for logged-out users

but hopefully this sets up a good foundation for those things later.

Note that this only affects "free" layouts - so e.g. online rounds and analysis layouts don't show any announcements. Is this desirable?

### DIFF
--- a/src/announce.ts
+++ b/src/announce.ts
@@ -1,0 +1,19 @@
+export interface Announcement {
+  msg: string
+  date: string
+}
+
+let announce: Announcement | undefined
+
+export function get(): Announcement | undefined {
+  return announce
+}
+
+export function set(a: Announcement | undefined): void {
+  announce = a
+}
+
+export default {
+  get,
+  set,
+}
--- a/src/announce.ts
+++ b/src/announce.ts
@@ -14,7 +14,14 @@ function keyOf(a: Announcement): string {
   return a.date // includes milliseconds, effectively unique
 }
 
+function isExpired(dateStr: string): boolean {
+  return new Date(dateStr) < new Date()
+}
+
 export function get(): Announcement | undefined {
+  if (announce && isExpired(announce.date)) {
+    announce = undefined
+  }
   return announce
 }
 
@@ -43,7 +50,7 @@ export async function dismiss(): Promise<void> {
 
   // clean up existing dismissed cache and add this new entry
   const existingDismissed = await asyncStorage.get<string[]>(STORAGE_KEY)
-  const newDismissed = (existingDismissed || []).filter(dateStr => (new Date(dateStr) >= new Date()))
+  const newDismissed = (existingDismissed || []).filter(dateStr => !isExpired(dateStr))
   newDismissed.push(dismissKey)
   await asyncStorage.set(STORAGE_KEY, newDismissed)
 
--- a/src/announce.ts
+++ b/src/announce.ts
@@ -1,4 +1,7 @@
+import asyncStorage from './asyncStorage'
 import redraw from './utils/redraw'
+
+const STORAGE_KEY = 'announce'
 
 export interface Announcement {
   msg: string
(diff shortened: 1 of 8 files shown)


veloce:
Thanks! I think that indeed this message should be on all layouts, game included.
It should be a toast like message: something that display above the UI (with some little transparency I guess), so without "moving" things on the screen.
I think we need a toast component in the app, since the system toast disappear to fast and is not visible enough.

src/i18n.ts
```diff
@@ -93,6 +94,13 @@ export function fromNow(date: Date): string {
   })
 }
 
+export function distanceToNowStrict(date: Date, addSuffix = false): string {
```
veloce, line 97: considering there is already a `formatRelative`, is this new function really needed?

src/session.ts
```diff
@@ -302,38 +304,37 @@ function rememberLogin(): Promise<Session> {
...
 
-function refresh(): Promise<void> {
-  return fetchJSON<Session>('/account/info', { cache: 'reload' })
-  .then((data: Session) => {
+async function refresh(): Promise<void> {
+  try {
+    const data = await fetchJSON<Session>('/account/info', { cache: 'reload' })
     session = data
     storeSession(data)
+    announce.set(data.announce)
```
veloce, line 312: You are thinking of an `announce` data here in case people would miss the websocket message? I don't know if it's necessary, I guess we should see that with @ornicar if needed.

If you want to start implementing the websocket handler here you can btw.

jas14: Yeah - on the desktop site, from what I understand, the `announce` data is injected on page load or sent once via websocket. My concern is that mobile users who are not online when the announcement is made will miss the websocket message, so they'll never know about it otherwise.

src/i18n.ts
jas14: There is a difference: `formatRelative`/`fromNow` shows the given date in words relative to now, e.g. "today at 4:30 AM" or "Sunday at 4:30 AM," where `distanceToNow[Strict]` basically displays the _interval_ between now and the given date, e.g. "in 12 hours" or "in 10 minutes". I chose `distanceToNow` as is what's used on desktop and it takes less screen space.

Let me know if you would prefer `formatRelative` though and I'll change it, it's an easy switch 👍

veloce: 👍

jas14:
> Thanks! I think that indeed this message should be on all layouts, game included.
> It should be a toast like message: something that display above the UI (with some little transparency I guess), so without "moving" things on the screen.

Hmm, I'm having a little trouble figuring out where to put this on the non-free layouts. (Or really, the `board` layout, since I don't think anyone would care to see the announcement while using the full-screen clock.) I can imagine a bullet or hyperbullet player being rather annoyed if a toast appeared on top of their game. I can also imagine someone putting down their phone for a minute or so and missing a timed toast. But the banner would shift the screen content.

--------------

What do you think of something like this: putting a high-visibility button in the nav bar that would show an announcement overlay when clicked? I know the nav bar is crowded, but 1) I don't expect announcements to be made often and 2) I could add some kind of "Acknowledge" button to the overlay, which would permanently hide that announcement.



I know

veloce:
Sorry for the delayed reply, I missed your last message.

Why not having, instead of the button, the message displayed over the navbar itself, in a high visibility background like your orange examples above?
There should be enough space here to display server restart announcements (which are the only ones that I know so far, and certainly the most importants). For longer texts we could have a button to expand and see the rest.

That way I think people will be warned but the game isn't disturbed.

src/announce.ts
```diff
@@ -0,0 +1,56 @@
...
+
+let announce: Announcement | undefined
+
+function keyOf(a: Announcement): string {
+  return a.date // includes milliseconds, effectively unique
+}
+
+export function get(): Announcement | undefined {
+  return announce
+}
```
jas14, line 19: TODO: ensure `announce` is not expired

jas14:
Updated! Users can tap to toggle announcement expansion, if applicable.

Long announcement, not expanded:


Long announcement, expanded:


Short announcement:

veloce:
That's great @jas14 
What is missing now before this can go to production?
Here is a first task list of what I'm thinking:

- [ ] show also in board layouts
- [ ] handle by WS and by calling `/account/info`; how would that work? is server API ready?
- [ ] test; how do you test this?

jas14:
> Show also in board layouts

Just pushed be0b715

> handle by WS and by calling `/account/info`; how would that work? is server API ready?

yep, server API is ready! see https://github.com/ornicar/lila/issues/8185. (and the server has already been sending announcement messages via websocket for quite some time, but the app currently ignores them.)

> test; how do you test this?

I gave one of my local accounts administrative permission, then went to /dev/cli on the website, and issued the `announce <time> <message>` command

- Websocket: I had a mobile client open on the home screen and on a game, and confirmed that the announcement appeared on both screens.
- `/account/info`: To force polling `/account/info`, I did each of the following and confirmed that the announcement appeared:
  - refreshed the page while on the home screen (logged in)
  - refreshed the page while in a game
  - logged into the app from a logged-out state
- Dismissing: I dismissed the announcement, then repeated the above `/account/info` steps (testing dismissal of announcements received via WS wouldn't make sense, as an announcement is only ever sent once via WS) and confirmed that the announcement no longer appeared. I made another announcement and observed that the second one did appear.
- Expiration: I made an announcement for 5 seconds in the future and confirmed that the announcement disappeared on redraw.
- Message expansion: I made a very long announcement, then confirmed that the message was ellipsized; then confirmed that tapping on the announcement element expands to show the full text, and tapping again collapses back to the original ellipsized state.

I'm happy to write some unit tests of the `announce` module as well! fwiw I don't see any automated existing DOM testing.

veloce:
So it looks pretty much ready :) I don't think unit tests are really useful for such a module. I'm gonna test this asap on stage, and merge it if all is ok.

PR #1565, opened 2021-02-21
Add i18n to variant performance stats

tmutas:
I took the liberty of adding existing translations from crowdin for the variant performance stats view.
I didn't find a proper translated context only for "streaks" and "percentile", so I left those, but otherwise I think it looks good.

It's my first pull request here, so just let me know if something's not quite right!


veloce:
Everything looks fine, thanks!

PR #1601, opened 2021-03-15
Set up repository for CrowdIn

jas14:
🌍  🌎  🌏 

This PR sets up configuration such that this repo can be added to the lichess CrowdIn project.

Once this PR is merged, a CrowdIn administrator can add this repo, selecting the master branch as the source branch and maybe `i18n-master` (or whatever name they want) as the dest branch. Then, any changes in JSON files in `translation/source/*.json` will be automatically imported into the CrowdIn project.

As strings are translated, CrowdIn will automatically commit changes into `translation/dest/<locale>/<filename>.json` on the `i18n-master` branch. When we're ready to bring said translations back into the master branch, we can create a merge PR.

Then, we run `npm run i18n-update`. This will download and unpack requested lila translation modules as before, but this time it will also add all mobile-specific CrowdIn strings (at `translation/dest/<locale>/<filename>.json`).

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,0 +1,46 @@
+#
+# Your Crowdin credentials
+#
+"project_id": ""
+"api_token": ""
+"base_path": "."
+"base_url": "https://api.crowdin.com"
+"commit_message": 'New translations: %original_file_name% (%language%)'
+"append_commit_message": false
+
+#
+# Choose file structure in Crowdin
+# e.g. true or false
+#
+"preserve_hierarchy": true
+
+#
+# Files configuration
+#
+files: [
+ {
+  #
+  # Source files filter
+  # e.g. "/resources/en/*.json"
+  #
+  "source" : "/translation/source/*.json",
+
+  #
+  # Where translations will be placed
+  # e.g. "/resources/%two_letters_code%/%original_file_name%"
+  #
+  "translation" : "/translation/dest/%locale%/%file_name%.%file_extension%",
+
+  #
+  # File type
+  # e.g. "json"
+  #
+  "type" : "json",
+
+  #
+  # The parameter "update_option" is optional. If it is not set, after the files update the translations for changed strings will be removed. Use to fix typos and for minor changes in the source strings
+  # e.g. "update_as_unapproved" or "update_without_changes"
+  #
+  "update_option" : "update_without_changes"
+ }
+]
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,8 +1,6 @@
 #
 # Your Crowdin credentials
 #
-"project_id": ""
-"api_token": ""
 "base_path": "."
 "base_url": "https://api.crowdin.com"
 "commit_message": 'New translations: %original_file_name% (%language%)'
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -29,6 +29,8 @@ files: [
   #
   "translation" : "/translation/dest/%locale%/%file_name%.%file_extension%",
 
+  "labels" : "mobile",
+
   #
   # File type
   # e.g. "json"
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -29,7 +29,7 @@ files: [
   #
   "translation" : "/translation/dest/%locale%/%file_name%.%file_extension%",
 
-  "labels" : "mobile",
+  "labels" : ["mobile"],
 
   #
   # File type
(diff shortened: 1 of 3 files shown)


veloce:
That's great! Now I think the untranslated string object should be reviewed because it's quite old and we probably don't want to translate everything in it. I'll put comments on specific lines.

translation/source/mobile-misc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "apiUnsupported": "Your version of lichess app is too old! Please upgrade for free to the latest version.",
+  "apiDeprecated": "Upgrade for free to the latest lichess app! Support for this version will be dropped on %s.",
+  "resourceNotFoundError": "Resource not found.",
+  "lichessIsUnavailableError": "lichess.org is temporarily down for maintenance.",
+  "lichessIsUnreachable": "lichess.org is unreachable.",
+  "mustSignInToJoin": "You must sign in to join it.",
+  "playerisInvitingYou": "%s is inviting you",
```
veloce, line 8: not sure we should keep translations for challenges; maybe there is a way to use the website ones.

translation/source/mobile-misc.json
```diff
@@ -0,0 +1,17 @@
...
+  "toATypeGame": "To a %s game",
+  "unsupportedVariant": "Variant %s is not supported in this version",
+  "notesSynchronizationHasFailed": "Notes synchronization with lichess has failed, please try later.",
+  "returnToHome": "Return to home",
+  "localEvalCaution": "Caution: intensive usage will drain battery.",
```
veloce, line 13: This warning should be removed completely, so let's not translate it.

translation/source/mobile-misc.json
```diff
@@ -0,0 +1,17 @@
...
+  "incorrectThreefoldClaim": "Incorrect threefold repetition claim.",
```
veloce, line 14: I bet the threefold offline claim is rarely used, I wouldn't translate it.

translation/source/mobile-misc.json
jas14: The website communicates the challenge almost entirely through iconography:
Screen Shot 2021-03-20 at 9 50 36 AM

We could easily get rid of the `toATypeGame` key by adopting the website's "Game Type · Time Control · Variant" string format - I can open a separate PR for that.

Re: the `playerisInvitingYou` key: I think the card would need to be reworked a little bit to get rid of it.
Screen Shot 2021-03-20 at 9 50 58 AM

Any suggestions?

veloce: I liked how you tweaked it in the other PR. So we keep only the `playerIsInvitingYou` key

jas14:
Comments addressed 👍

crowdin.yml
```diff
@@ -0,0 +1,46 @@
...
+  #
+  "translation" : "/translation/dest/%locale%/%file_name%.%file_extension%",
+
+  "labels" : ["mobile"],
+
+  #
+  # File type
+  # e.g. "json"
+  #
+  "type" : "json",
```
niklasf, line 38: At the time we chose xml for lila, because it was the simplest format with support for plural forms. No plurals needed in lichobile, or does json support plurals, now?

jas14: Good question. We've had to [flatten plurals](https://github.com/veloce/lichobile/blob/ae09c60192bffdbe1161af5c430afedacf2893dd/scripts/i18n-update.ts#L118-L122) from the lila XML anyway, so same difference on the lichobile side. It'll just end up being [multiple keys](https://github.com/veloce/lichobile/blob/ae09c60192bffdbe1161af5c430afedacf2893dd/src/i18n.ts#L204).

Looks like Crowdin consolidates editing plurals grouped under the same key, so if that's desirable over totally separate keys, I can switch this to XML

jas14: In fact... we could even switch to [yaml](https://support.crowdin.com/file-formats/yaml/), which I find significantly more concise than XML...

niklasf: Looks like yaml would work. too. The reason it can't be flattened in the source, only the translation, is that some planguages have more plural forms than others. Example:

https://github.com/ornicar/lila/blob/ca3970720ef808800b7b3b72334da345d38fe8bd/translation/dest/site/ru-RU.xml#L37-L42

So `pluralKey` (https://github.com/veloce/lichobile/blob/b76cccb0d6f7faa478ee1072c8abdbf4c850299e/src/i18n.ts#L35) can be more than just `...:one` and `...:other`, depending on the language.

jas14: #1618 up to switch to YAML

PR #1608, opened 2021-03-19
Fix first move check aura not appearing in puzzles

### LINKED ISSUE #1597: King in a puzzle check is not red.
Screenshot_20210314-154018_lichess
Being checked in a puzzle doesn't turn the king red

**To Reproduce**
1. Go to 'Lichess Mobile'
2. Click on 'Puzzles'
3. Wait for a puzzle that checks you.
4. Bug found

**Expected behavior**
All checked kings are red with the given setting enabled.

**Screenshots**

 - Device: Galaxy A12.
 - OS: unsure. Latest.
 - Version: Latest.

jas14:
As the server [doesn't send the "check" property](https://github.com/ornicar/lila/blob/a075a5f105e2333916e3f28e0b39aa52e5520170/modules/puzzle/src/main/GameJson.scala#L111-L115), we need to infer it from the first move SAN. Closes #1597.

### DIFF
--- a/src/ui/training/TrainingCtrl.ts
+++ b/src/ui/training/TrainingCtrl.ts
@@ -287,7 +287,7 @@ export default class TrainingCtrl implements PromotingInterface {
       orientation: this.data.puzzle.color,
       movableColor: this.gameOver() ? null : this.data.puzzle.color,
       dests: dests || null,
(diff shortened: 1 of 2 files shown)


veloce:
Thank you!

PR #1610, opened 2021-03-20
Tweak challenge card

jas14:
To remove `toATypeGame` i18n key, as mentioned in https://github.com/veloce/lichobile/pull/1601#discussion_r598081391. Ordering (time - variant - rating mode) is consistent with the quick-pairing overlay:
Screen Shot 2021-03-20 at 11 15 21 AM


veloce:
looks good!

PR #1613, opened 2021-03-21
Set check when receiving castle move from API

### LINKED ISSUE #1605: When castling results in a check, king square does not turn red like it usually does
**Describe the bug**
If I castle and that puts the opponent king in check, the square doesn't turn red like it usually does when in check

**To Reproduce**
1. Example game: https://lichess.org/ydzTw40T
2. When I review the game, the square does turn red. But when playing the first time, it doesn't. See screenshot below

**Expected behavior**
Usually the king square goes red when the king is in check

**Screenshots**
IMG_5853

 - Device: iPhone SE
 - OS: iOS 14.4
 - Version v7.9.0

jas14:
Currently, when applying a move from the server, we [handle castles differently](https://github.com/veloce/lichobile/blob/c1c28216880984bad756a35a2b50381299a4dfd1/src/ui/shared/round/OnlineRound.ts#L483-L494) than other moves. This causes us to drop the `check` property when applying castles.

This PR fixes that by always applying the `check` property, regardless of whether the move is a castle or not. Closes #1605.

**Note**: as shown in the video below, this does _not_ cause the red check aura to appear when the move is applied locally (in this case, before move confirmation) - only when it comes from the server. This is currently the case whether the checking move is a castle or not. Users with high latency may continue to experience check aura appearance lag.

https://user-images.githubusercontent.com/569991/111905944-9d73cd80-8a24-11eb-936e-0528d9fa76cf.mov

### DIFF
--- a/src/ui/shared/round/OnlineRound.ts
+++ b/src/ui/shared/round/OnlineRound.ts
@@ -396,7 +396,7 @@ export default class OnlineRound implements OnlineRoundInterface {
     }
   }
 
-  public apiMove(o: MoveOrDrop) {
+  public apiMove(o: MoveOrDrop): void {
     const d = this.data
     const playing = gameApi.isPlayerPlaying(d)
 
@@ -501,6 +501,10 @@ export default class OnlineRound implements OnlineRoundInterface {
           chessFormat.uciToDropPos(o.uci),
           newConf
         )
+      }
+
+      if (o.check) {
+        this.chessground.setCheck(true)
       }
 
       if (o.promotion) {


veloce:
Thanks for fixing this. I made a slightly different fix here though: https://github.com/veloce/lichobile/commit/1bf4a10e21458b090bad5ba755776b4316c6e92d

The reason is that the `newConf` object was meant to be always set on chessground. The regression was introduced by a new condition to handle autocastling for chess960, and an optimisation to not re-render chessground twice during an api move (dangerous and probably not necessary optim...).

Issue #1617, opened 2021-03-22
Captured pieces appear backward


clementcontet:
**Describe the bug**
When playing offline "over the board", the pieces captured by the opponent are displayed backward

**To Reproduce**
1. Go to 'Play offline > Over the board'
2. Capture some pieces
3. See error

**Expected behavior**
Pieces captured by the opponent should be aligned with the direction of the text.

**Screenshots**
![backward piece](https://user-images.githubusercontent.com/7261426/111963604-00806580-8af4-11eb-90ac-8105c9ea293a.jpg)

**Smartphone (please complete the following information):**
 - OS: Android
 - Version: 7.9.0

PR #1622, opened 2021-03-23
i18n a couple OTB settings

jas14:
Ran into this while working on #1617, and now that mobile i18n is set up, we can fix it 🎉


veloce:
Need to wait for the translations to be pulled in `www` folder before merging it I think.

jas14:
> Need to wait for the translations to be pulled in www folder before merging it I think.

I think just the `en-US` "translations" should suffice, as we fall back on those - that is, it won't be any worse than now. The sources won't be synced to Crowdin until we merge into `master` anyway.

PR #1633, opened 2021-03-31
Option to choose board side in landscape mode

### LINKED ISSUE #1353: option required for chess board on Right side of screen.
It's more convenient for the board to be on the left side if you're left-handed, and it's already there.

_Originally posted by @ornicar in https://github.com/veloce/lichobile/issues/1343#issuecomment-677920648_

Ornicar;  
Hi;   By your logic its more convenient to be on the left side if left handed.  70-90 % of people are RIGHT HANDED.  so the app is less convenient for 70-90% of your tablet users.  Create the option is the root issue request.  Your opinion of convenience is irrelevant, an option is required. 

Thank you kindly;

### LINKED ISSUE #1541: IPad: Option to switch Board and UI Panel view.
Problem:
When I play on the IPad I always have the board on the left and the UI Panels (with the timer, chat and resign buttons etc.) on the right. This is very annoying when doing speed-chess, because my right hand is covering up the timer the notification when my opponent either writes a message, resigns or offers a draw.
As I always hover with my hand over the board, to prepare a fast move, my hand is constantly covering up the timer.
I flagged many times because of this.

Solution:
I would like to have an option in the settings to simply switch the board and the UI panels view. You could call it "left-hand" or "right-hand" mode. I think the current layout is even for left-handers (I am right-handed).

jas14:
Closes #1353. Closes #1541.

The setting:

Offline round:

Analysis:

Online round:

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -136,6 +136,7 @@ export default {
     pieceMove: prop<'tap' | 'drag' | 'both'>('game.pieceMove', 'both'),
     rookCastle: prop<0 | 1>('game.rookCastle', 1),
     moveList: prop<boolean>('game.moveList', true),
+    landscapeBoardSide: prop<'right' | 'left'>('game.landscapeBoardSide', 'left'),
   },
 
   analyse: {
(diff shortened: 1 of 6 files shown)


veloce (approved):
Code is ok, I just need the conflicts to be solved before merging. Thanks!

PR #1656, opened 2021-04-22
Add an intent filter for receiving PGN's from other apps

alexbakker:
This allows other apps to share PGN's with Lichess through an intent. See the following video for a demo with the Chess.com app: https://alexbakker.me/u/gbqja44mbi.mp4.

I don't have any experience writing TypeScript, so any feedback is welcome.


veloce:
Thanks, it looks good!

Now, having the same feature on iOS would be nice too. I wonder if it could work with url schemes?

veloce:
I just tested chesscom iOS app and apparently the way to go is to copy/paste the PGN, so no need to worry about iOS actually.

alexbakker:
Cool. Let me know if you need anything else from my end before this can be merged.

alexbakker:
What's the status here?

PR #1663, opened 2021-04-29
remove `booster` field

kraktus:
I could not test these changes since I do not have a mobile env.

I didn't remove the i18n key `thisPlayerArtificiallyIncreasesTheirRating` since I assume will be erased by next crowdin dump.


veloce:
There is a syntax error in one file as shown in the checks. Just for info, you can test your changes to some extent without a mobile env. The readme gives you the way to run the app in a browser.

PR #1675, opened 2021-05-08
Messages: replace instead of push state to match web

### LINKED ISSUE #1659: Buggy inbox in lichess Mobile App
**Describe the bug**
If you start a chat using the lichess mobile app, it can be buggy to return to the Mainpage if you have received new messages. for example, if i receive two new messages from the same person, then i will have to click the "go back" arrow two times in order to return to the main page.

**To Reproduce**
1. Go to your inbox and start a conversation with someone
2. Both user should send multiple messages
3. try to go back to the main menu.
4. You'll notice that going to back to the main menu is not so smooth. 

**Expected behavior**
I should be able to return to the main menu with a maximum of two clicks, the new messages should automatically show up on my inbox without needing to refresh.

 - Device: [e.g. iPhone xs]
 - OS: [e.g. Most recent verison]

jas14:
May fix #1659. The mobile app used `History.pushState` where the website used `History.replaceState`, which made it difficult to navigate out of chat using the native back button after opening conversations many times.

After this PR, the messages behavior will be as follows:
- When opening a conversation, the current history entry, `/inbox`, will be replaced with the conversation path (`/inbox/<userId>`).  
- When navigating back to the inbox via the back arrow, the history entry will not be replaced (that is, the URL will reflect the specific conversation that was just closed).

This means that at any point in the messages UI, hitting the native "back" button will exit the messages UI entirely. I don't love this behavior, but 1) it matches the website and 2) I see no way to avoid it without relying on `history.back()`, which would cause an unnecessary XHR reload of the inbox.

Tested locally using Chrome.

### DIFF
--- a/src/ui/msg/MsgCtrl.ts
+++ b/src/ui/msg/MsgCtrl.ts
@@ -40,7 +40,7 @@ export default class MsgCtrl {
       this.search.result = undefined
       this.loading = false
       if (data.convo) {
-        router.History.pushState(`/inbox/${data.convo.user.name}`)
+        router.History.replaceState(undefined, `/inbox/${data.convo.user.name}`)
         this.onLoadConvo(data.convo)
         redraw()
       }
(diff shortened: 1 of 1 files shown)


veloce:
Thanks! I need to test this in order to understand exactly what's going on here.

PR #1677, opened 2021-05-10
Fix piece color for past best move when promo

### LINKED ISSUE #1676: Pawn promotion suggestion in Analysis board changes color when it's the opposite color's turn to move
I noticed the bug while analyzing the puzzle 94449327.

The below bug occurs at [move 40 of this game](https://lichess.org/9sVANEDz/black#79) in the Lichess mobile Android app.
The promotion suggestion which is a black Queen here changes to White Queen when it's white's turn to move in the app. 
I don't see the same behavior in the Firefox browser on mobile or PC.

Observed behavior:
Pawn promotion suggestion in Analysis board changes color when it's the opposite color's turn to move.
Expected behavior:
Pawn promotion suggestion in Analysis board should stay the same color.

Phone model: OnePlus 6
OS: OxygenOS 10.3.9 / Android version 10

Attached snapshots.

Suggestion shows a Black Queen on Black's move 40 as seen in this snapshot:
Screenshot_20210509-234810
Suggestion changes to White Queen on white's move 41 as seen in this snapshot:
Screenshot_20210509-234830

jas14:
Closes #1676.

Tested with a local client against [the game mentioned](https://lichess.org/9sVANEDz/black#79) in the issue.

I also tried to test against a [random bughouse game](https://lichess.org/Jf7wjgRf/black#19) to verify drop move suggestions are rendered as expected. However, it appears the server doesn't provide best move suggestions for drops, which is when this code path would be exercised.

### DIFF
--- a/src/ui/analyse/view/boardView.ts
+++ b/src/ui/analyse/view/boardView.ts
@@ -8,6 +8,7 @@ import { Shape } from '../../shared/BoardBrush'
 import Clock from './Clock'
 import { povDiff } from '../ceval/winningChances'
 import AnalyseCtrl from '../AnalyseCtrl'
+import { opposite } from '~/chessground/util'
 
 export default function renderBoard(ctrl: AnalyseCtrl) {
   return h('div.analyse-boardWrapper', [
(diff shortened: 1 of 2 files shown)


src/ui/analyse/view/boardView.ts
```diff
@@ -8,6 +8,7 @@ import { Shape } from '../../shared/BoardBrush'
 import Clock from './Clock'
 import { povDiff } from '../ceval/winningChances'
 import AnalyseCtrl from '../AnalyseCtrl'
+import { opposite } from '~/chessground/util'
```
jas14, line 11: chessops appears to have an identical `opposite` function - should we use that here? and if so, should we replace all instances of chessground/util opposite witth chessops opposite?

Issue #1416, opened 2020-11-14
Lichess android app


amermiftari:
the android app is not  full screen for some devices that have 18.5x9 aspect ratio and some other screens 2x1 ratio

Issue #1145, opened 2020-02-14
App does not fill whole screen (samsung S10e)


maelic13:
# Bug report

**Current behavior:**

The lichess mobile app, both version 6.3.0 from play store and 7.0.1 downloaded from here, do not contain the whole screeen of my phone, there is a black bar at the bottom. My resolution is 1080 x 2280 pixels, 19:9 ratio, and am running on ANdroid 10.

**Expected behavior:**

Application should contain the whole screen to make use of as much space as it is given.

**Environment information**

- phone hardware: https://www.gsmarena.com/samsung_galaxy_s10e-9537.php
- phone operating system version: Android 10
- version of lichess mobile: both 6.3.0 from play store and 7.0.1 from github

**Other information:**

I do not remember very well, but I think before upgrade to android 10 from android 9, the bug was not there. But I cannot say for sure and also cannot test it now..

![Screenshot_20200214-072306_lichess](https://user-images.githubusercontent.com/30387931/74506963-6666ea80-4efb-11ea-9c45-85fc6c947c76.jpg)

maelic13:
> Same thing was happening to me on my Samsung S10 on 6.3.0 and the current 7.0.4, but I found that you can fix it by going to Settings > Display > Full Screen Apps > Lichess > Set to full screen instead of auto.

Indeed that works! Thank you.

MarionMoseby:
I am experiencing the same bug, and the fix more or less does apply. Will this get fixed?

PR #1685, opened 2021-05-15
Fix black bars visible on Android devices with narrow aspect ratios

cvb941:
This fixes black bars visible at the top and bottom of the screen on some devices (like Samsung S8, etc.), as mentioned in #1145 and #1416.

If the `android:resizeableActivity` flag was set to `false` on purpose for some reason, we can also set the [minimum aspect ratio](https://developer.android.com/guide/practices/screens-distribution#MaxAspectRatio) instead to 2.1.

Also, the `android:resizeableActivity` flag can be removed altogether now, as it is true by default.


veloce:
I could not reproduce the black bars issue on a Samsung S10 emulator (from genymotion). Perhaps a real device is needed.

Issue #1691, opened 2021-05-21
Show previous score on coordinate training


NeonDactyl:
**Is your feature request related to a problem? Please describe.**
When I do coordinate training on mobile, my thumb position usually covers the score at the bottom, and it's frustrating to not see my score after the run.

**Describe the solution you'd like**
I'd like to see a small addition above the average scores to show the previous score.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1692, opened 2021-05-21
#1691 - added functionality desired for previous score shown in coordinate training

NeonDactyl:
Added simple functionality and two translations, more translations will be needed.

### DIFF
--- a/www/i18n/en-US.js
+++ b/www/i18n/en-US.js
@@ -1099,6 +1099,7 @@ export default {
   "playedXTimes:other": "Played %s times",
   "coordinates": "Coordinates",
   "coordinateTraining": "Coordinate training",
+  "previousScore": "Previous score: %s",
   "averageScoreAsWhiteX": "Average score as white: %s",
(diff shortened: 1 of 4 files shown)


www/i18n/en-US.js
```diff
@@ -1099,6 +1099,7 @@ export default {
   "playedXTimes:other": "Played %s times",
   "coordinates": "Coordinates",
   "coordinateTraining": "Coordinate training",
+  "previousScore": "Previous score: %s",
```
veloce, line 1102: Please don't modify these files: they're auto generated. Just put an english string for now where you need it, we'll consider translating it later.

veloce (changes requested):

www/i18n/en-US.js
NeonDactyl: > Please don't modify these files: they're auto generated. Just put an english string for now where you need it, we'll consider translating it later.

Apologies, I didn't realize they were generated, but they are reverted now.

I'm interested to learn how that's accomplished.

PR #1732, opened 2021-06-18
Add button to share game GIF

### LINKED ISSUE #1162: Feature request: add "share as gif option in app
Just like the web version, add share the game as gif option in the app too.

photo_2020-03-24_13-54-29
photo_2020-03-24_13-54-44

jas14:
From analyse control, like the website. Closes #1162. It's only the GIF URL, not the GIF data itself, but the latter would require a lot more effort.

Screen Shot 2021-06-18 at 5 35 05 PM

### DIFF
--- a/src/lichess/game.ts
+++ b/src/lichess/game.ts
@@ -125,7 +125,7 @@ export function nbMoves(data: OnlineGameData, color: Color) {
   return Math.floor((data.game.turns + (color === 'white' ? 1 : 0)) / 2)
 }
 
-export function result(data: GameData) {
+export function result(data: GameData): string {
   if (gameStatus.finished(data)) switch (data.game.winner) {
     case 'white':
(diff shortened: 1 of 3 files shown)


Aditya-ds-1806:
Can we have these options when we open a game to just view it? Why go to the analysis and share from there? Adds an unnecessary step.

jas14:
Generally, the less frequent/common an action is, the more clicks it's acceptable to hide the feature behind; this is not only because it doesn't add much work over time to get to the infrequent action, but because it would otherwise add more cognitive work to the flow of accessing frequent/common actions, as you'd have to spend more time searching for the correct one.

Specifically here, going to the analysis screen is only one additional step. You'll notice that the website actually hides this behind _two_ additional steps: "go to analysis" -> "go to the FEN + PGN tab."

PR #1747, opened 2021-06-30
Enable type-based linter rules

### LINKED ISSUE #1744: Enable lint rule: @typescript-eslint/strict-boolean-expressions
We have [a bug](https://github.com/veloce/lichobile/issues/1727) probably caused by using non-booleans in conditionals. Let's enable [@typescript-eslint/strict-boolean-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md), with options as necessary, to avoid these in the future.

jas14:
And upgrade NPM. Closes #1744.


jas14:
https://github.com/npm/cli/issues/2610

src/chessground/drag.ts
```diff
@@ -178,7 +178,7 @@ export function end(ctrl: Chessground, e: TouchEvent) {
   // board editor mode: delete any piece dropped off the board
   else if (cur.started && draggable.deleteOnDropOff) {
     state.pieces.delete(cur.orig)
-    setTimeout(state.events.change || util.noop, 0)
+    setTimeout(() => (state.events.change || util.noop)(), 0)
```
veloce, line 181: why this change?

jas14: https://eslint.org/docs/rules/no-implied-eval

veloce: but I don't get it, as here this is not a string argument that is passed to `setTimeout`, it is a function. I'm not a fan of this change really.

jas14: I think you're right, it's a bug in the rule... the type of the expression before this change is `() => void`, and you can "fix" the error by assigning the expression to a const first
```typescript
const fn = state.events.change || util.noop
setTimeout(fn, 0)
```

veloce: then I much prefer to have this instead of the arrow function wrapper

jas14: https://github.com/typescript-eslint/typescript-eslint/issues/3621

jas14: Just backed out this change, we can re-enable the rule later once the bug is fixed

Issue #1366, opened 2020-09-08
iOS loading screen in dark mode


tmmlaarhoven:
When loading the app, briefly a loading screen appears with the lichess logo in black and white: a black lichess knight on a white background. When using dark mode however, wouldn't it make sense to use a negative of that, with a white knight on a black background? (It would be easier on the eyes at night, and more in line with the dark mode.)

**Is your feature request related to a problem? Please describe.**
Not so much a problem, but I like things to be dark on my phone, and not be blinded by light when opening the app at night.

**Describe the solution you'd like**
Change the loading screen for users in dark mode from the current one to its negative.

**Describe alternatives you've considered**
Not applicable. 

**Additional context**
I think the description above is clear enough.

tmmlaarhoven:
Illustrative before and after images:

![Screenshot_20200908-234125](https://user-images.githubusercontent.com/11218275/92541515-8a4a9e80-f246-11ea-9c35-655c85f66076.jpg)

![image](https://user-images.githubusercontent.com/11218275/92541517-8ae33500-f246-11ea-9590-9766c1dc43a9.png)

jas14:
Should be natively doable in Android and iOS, relevant Capacitor issue here: https://github.com/ionic-team/capacitor/issues/2947

DaniSpringer:
Should not be hard. If I push such changes will you accept them? (I can do iOS)

jas14:
@DaniSpringer yes, PRs are always welcome! Thanks

DaniSpringer:
> @DaniSpringer yes, PRs are always welcome! Thanks

Hi, so I now notice that the app isn’t native? I only know pure swift + Xcode

jas14:
Yes, the app is written mostly in cross-platform JS, using Capacitor as a framework. There are iOS and Android subdirectories for platform-specific things, though. You might start with the above-linked Capacitor issue for inspiration

DaniSpringer:
> Yes, the app is written mostly in cross-platform JS, using Capacitor as a framework. There are iOS and Android subdirectories for platform-specific things, though. You might start with the above-linked Capacitor issue for inspiration

Any plans to make a native app? I believe I would be more helpful then. I don’t have the budget to learn new technologies after-hours at the moment, unfortunately.

jas14:
Not to my knowledge

ncodee:
Any reason why this hasn't been implemented yet? Surely, it should be a simple change.

[...]

PR #1749, opened 2021-07-02
Android dark launch

jas14:
The people demand it!
Part of #1366.

Tested locally on a Pixel 4 API 29 emulator.
Did not test on any API <23 devices.

https://user-images.githubusercontent.com/569991/124998653-cb544080-e01a-11eb-83d7-8f34d0af1005.mov


veloce:
Indeed the white flash is weird, as the background color should be the one defined here: https://github.com/veloce/lichobile/blob/master/capacitor.config.json#L7

jas14:
Oh.. good point, I think we want `#000000FF`.

https://capacitorjs.com/docs/apis/splash-screen#background-color

> Possible values for backgroundColor are either `#RRGGBB` or `#RRGGBBAA`.

jas14:
Fixed! Video updated to reflect the current code.

veloce:
The problem we might have now, is a black flash when you load the app in light mode. Have you checked that?

src/main.ts
```diff
@@ -36,4 +36,8 @@ settingsInit()
 .then(themeInit)
 .then(i18nInit)
 .then(() => processWindowLocation())
-.then(() => Plugins.SplashScreen.hide())
+.then(() => {
+  setTimeout(() => {
```
veloce, line 40: why adding this timeout?

jas14: I [searched](https://www.google.com/search?q=capacitor+white+screen+after+splash) for this problem and several people recommended setting a small timeout to avoid the flash. This should mitigate a black flash in light mode - I'll upload a screen capture

veloce: yes I guess it should be fine like that; and it's better to have a black background than a white one.

jas14: https://user-images.githubusercontent.com/569991/125131051-1d0ad280-e0d0-11eb-9e3b-f91b7e53f953.mov

veloce:
This is great! Thanks. Now we need to do the same on iOS, so I'm keeping this PR open in order not to forget to do it. I might give it a try this weekend.

veloce:
@jas14 I tested on my device (nokia 5, android 9) and it doesn't work, I always get the white launcher screen.

jas14:
@veloce do you mean you still see the white flash, or the whole launcher is the usual "light" one? If the latter, are you using the device's dark mode, or just a dark theme in the app?

veloce:
I see the usual light splash screen, and I'm using device's dark mode and dark theme in app.

jas14:
@veloce this seems to be a device-specific problem: https://community.phones.nokia.com/discussion/57874/dark-mode-theme-not-properly-working

jas14:
> Now we need to do the same on iOS

@veloce #1783 is ready 🙂

veloce:
> @veloce this seems to be a device-specific problem: https://community.phones.nokia.com/discussion/57874/dark-mode-theme-not-properly-working

ok I'll test with another device and if it works, I'll ship it as is. No big deal if it doesn't work on all devices. As far as I understand, the dark launch theme would appear only when you set your **os** theme to dark, right? (assuming that the app settings has not impact on it).

jas14:
That's correct, the launch dark/light color is only affected by the OS theme, not the app theme.

PR #1760, opened 2021-07-06
Push instead of replace path when going to analysis

### LINKED ISSUE #1758: Enter and exit analysis twice affects the current game - it gets closed and changed to the previous one
**Describe the bug**
The current game is closed without any warning after twice entering after-game analyses and exiting it with back button.

**To Reproduce**
1. Play games without account (maybe with account too, don't know)
2. Finish the game
3. Click Analysis
4. Press Back button to exit analysis and get back to final position of the game
5. Click Analysis again
6. Clock Back button again
7. The game is lost (never can find the id) and the current position is from the previous game.

**Expected behavior**
Being in the game that was last played

 - Device: Android
 - OS: Android 8.1
 - Browser: Lichess mobile official application

**Additional context**
If one presses back in the usual situation - the Confirmation message is shown: _Do you really want to leave the game? You can't go back to it after_. But in this game - the current game is lost and closed without any confirmation. Bug for sure.

jas14:
Currently, the user can navigate back to the game round view from a _playable_ game by pressing the back arrow or the native back button; however, from a _non-playable_ game, the only way to do the same is via the back arrow. Both states should allow the use of the native back button to navigate back.

This `replace` parameter looks left over from a WIP commit from 2 years ago, so do correct me if I'm missing some more context here 🙂 

Closes #1758.

### DIFF
--- a/src/ui/shared/round/OnlineRound.ts
+++ b/src/ui/shared/round/OnlineRound.ts
@@ -74,8 +74,6 @@ export default class OnlineRound implements OnlineRoundInterface {
   private clockIntervId!: number
   private blur: boolean
 
-  private readonly playableOnInit: boolean
-
   private transientMove: TransientMove
   private appStateListener: PluginListenerHandle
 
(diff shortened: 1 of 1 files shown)


veloce:
I really don't remember here :)

I'll try to reproduce the steps described in the link issue to check that.

PR #1773, opened 2021-07-18
fix chat mute between SB players

ornicar:
A SB player must not see lines of other SB players, only their own

### DIFF
--- a/src/ui/shared/chat.ts
+++ b/src/ui/shared/chat.ts
@@ -84,7 +84,7 @@ export class Chat {
   // --
 
   private isLegitMsg = (msg: ChatMsg) => {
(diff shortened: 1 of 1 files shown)


src/ui/shared/chat.ts
```diff
@@ -84,7 +84,7 @@ export class Chat {
   // --
 
   private isLegitMsg = (msg: ChatMsg) => {
-    return !msg.d && (!msg.r || this.isShadowban) && !isSpam(msg.t)
+    return !msg.d && (!msg.r || msg.u == this.player?.user?.username) && !isSpam(msg.t);
```
jas14, line 87:
```suggestion
    return !msg.d && (!msg.r || msg.u === this.player?.user?.username) && !isSpam(msg.t)
```

PR #1818, opened 2021-08-24
Tournament player titles

### LINKED ISSUE #1814: Fide title display next to username in tournaments
I think that it will be great in mobile app if in tournaments i can see the fide title of a user next to the username.

jas14:
Closes #1814. Compare screenshot to the one in the issue.

### DIFF
--- a/src/ui/tournament/detail/tournamentView.tsx
+++ b/src/ui/tournament/detail/tournamentView.tsx
@@ -274,6 +274,14 @@ function renderNavButton(icon: string, isEnabled: boolean, action: () => void) {
   })
 }
 
+function renderPlayerTitle(player: {title?: string}): Mithril.Child {
+  if (player.title == null) {
+    return null
+  }
+
+  return h('span.userTitle', [player.title, h.trust('&nbsp;')])
+}
+
 function renderPlayerEntry(userName: string, player: StandingPlayer, i: number, teamColor?: number, teamName?: string) {
   const evenOrOdd = i % 2 === 0 ? 'even' : 'odd'
   const isMe = player.name === userName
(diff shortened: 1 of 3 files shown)


Baikalcadillac:
Small detail I’ve noticed - website version have FIDE titles below winner’s cup as well. 

4491E91E-258A-4398-8A65-68FE1B7FE9F7

jas14:
Thanks for the suggestion @Baikalcadillac , updated!

src/ui/tournament/detail/tournamentView.tsx
```diff
@@ -274,23 +274,34 @@ function renderNavButton(icon: string, isEnabled: boolean, action: () => void) {
...
+          {`${player.name} (${player.rating})`}
+        </span>
+        <span className={`playerTeam ttc-${ttc}`}> {teamName ?? ''} </span>
       </div>
       <div className={'tournamentPoints ' + (player.sheet.fire ? 'on-fire' : 'off-fire')} data-icon="Q">
         {player.score}
       </div>
     </li>
-  )
+  ) as Mithril.Child
```
veloce, line 304: Why casting here? It should be not necessary. It's preferable to put the function return type explicitly then.

jas14: Typescript doesn't seems to recognize JSX as Mithril.Vnode / Mithril.Child. The cast was to remove the "`any` return type" lint warning. A better long-term solution will be to tell Typescript about JSX or convert it all to `h` calls but that's a different PR :)

veloce: ah ok thanks. I didn't notice this warning actually

PR #1831, opened 2021-09-01
Fix promotion

### LINKED ISSUE #1802: antichess always promotes to queen
**Describe the bug**

When playing antichess against the computer, the computer always promotes a pawn to a queen, even when it is not optimal.

**To Reproduce**
1. Go to board editor
2. Set up the following FEN: `8/6k1/8/8/8/K7/3p4/2R5 b -- 1 2`
3. Continue from here
4. Play offline against Computer
5. Play White in Antichess

Computer promotes pawn to queen, immediately losing. I believe a rook promotion will lead to a draw.

**Expected behavior**
Computer should consider promoting to pieces other than a queen.

 - Device: Motorola One 5G Ace
 - OS: Android 10
 - Version Lichess v7.11.3

shabbychef:
Promotion piece received from engine now passed forward. This should close #1802.
In testing, it seems that the proper variant string had not been passed on. This is perhaps a separate issue, in commit 3050342.

### DIFF
--- a/src/stockfish.ts
+++ b/src/stockfish.ts
@@ -65,8 +65,6 @@ export class StockfishPlugin {
     if (this.isVariant()) {
       if (this.variant === 'threeCheck')
         return this.setOption('UCI_Variant', '3check')
-      else if (this.variant === 'antichess')
-        return this.setOption('UCI_Variant', 'giveaway')
       else
         return this.setOption('UCI_Variant', this.variant.toLowerCase())
     } else {
--- a/src/stockfish.ts
+++ b/src/stockfish.ts
@@ -65,6 +65,8 @@ export class StockfishPlugin {
     if (this.isVariant()) {
       if (this.variant === 'threeCheck')
         return this.setOption('UCI_Variant', '3check')
+      else if (this.variant === 'antichess')
+        return this.setOption('UCI_Variant', 'giveaway')
       else
         return this.setOption('UCI_Variant', this.variant.toLowerCase())
     } else {
--- a/src/ui/ai/AiRound.ts
+++ b/src/ui/ai/AiRound.ts
@@ -28,7 +28,6 @@ import Replay from '../shared/offlineRound/Replay'
 import actions, { AiActionsCtrl } from './actions'
 import Engine from './engine'
 import newGameMenu, { NewAiGameCtrl } from './newAiGame'
-import { charToRole } from 'chessops/util'
 
 interface InitPayload {
   variant: VariantKey
@@ -226,16 +225,10 @@ export default class AiRound implements AiRoundInterface, PromotingInterface {
   public onEngineMove = (bestmove: string) => {
     const from = <Key>bestmove.slice(0, 2)
     const to = <Key>bestmove.slice(2, 4)
+    const role = <Role>chessFormat.uciToProm(bestmove)
     this.vm.engineSearching = false
-    // does the promotion piece have to be passed to this function as well?
     this.chessground.apiMove(from, to)
-    if (bestmove.length > 4) {
-      const role = charToRole(bestmove.slice(4,5))
-      this.replay.addMove(from, to, role)
-    } else {
-      // inelegant to have this switch. Is there an empty Role to pass instead?
-      this.replay.addMove(from, to)
-    }
+		this.replay.addMove(from, to, role)
     redraw()
   }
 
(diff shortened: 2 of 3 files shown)


src/stockfish.ts
```diff
@@ -65,6 +65,8 @@ export class StockfishPlugin {
     if (this.isVariant()) {
       if (this.variant === 'threeCheck')
         return this.setOption('UCI_Variant', '3check')
+      else if (this.variant === 'antichess')
+        return this.setOption('UCI_Variant', 'giveaway')
```
jas14, line 69: @shabbychef looking at the lichess website's Stockfish protocol, I see stockfish.js (which is used only by the web plugin in this repository!) **must** [be passed `giveaway`](https://github.com/ornicar/lila/commit/c4be12ee3afc8332c4f4093d2c2fb86696d7fc98) instead of `antichess`. I believe the native implementations provided by capacitor-stockfish / capacitor-stockfish-variants use https://github.com/ddugovic/Stockfish. After digging around that code a little, it appears "giveaway" is a subvariant of "anti," though I can't find any documentation as to how the two differ.

Anyway, long story short, I'd probably keep the "antichess" string in general and only patch the web behavior, e.g.:
```suggestion
      else if (Capacitor.platform === 'web' && this.variant === 'antichess')
        return this.setOption('UCI_Variant', 'giveaway')
```

src/ui/ai/AiRound.ts
```diff
@@ -225,9 +225,10 @@ export default class AiRound implements AiRoundInterface, PromotingInterface {
   public onEngineMove = (bestmove: string) => {
     const from = <Key>bestmove.slice(0, 2)
     const to = <Key>bestmove.slice(2, 4)
+    const role = <Role>chessFormat.uciToProm(bestmove)
```
jas14, line 228: No need to cast here, in fact it's technically incorrect (could be `undefined` if the UCI isn't a promo move, which is OK for the `addMove` call)
```suggestion
    const role = chessFormat.uciToProm(bestmove)
```

src/ui/ai/AiRound.ts
```diff
@@ -225,9 +225,10 @@ export default class AiRound implements AiRoundInterface, PromotingInterface {
...
     this.vm.engineSearching = false
     this.chessground.apiMove(from, to)
```
jas14, line 230: I believe you'll need to conditionally call `this.chessground.promote(to, role)` if `role` is not undefined

jas14:
> Actually, it would seem that stockfish engine will expect `threecheck`, so line 67 is silently broken?

The native plugin [expects `3check`](https://github.com/ddugovic/Stockfish/blob/4e875080ad0f20a93cbae245d6e7c3d7d9a62c50/src/types.h#L225-L227). I believe stockfish.js expects the same. Where did you see `threecheck`?

src/stockfish.ts
shabbychef: got it, I did not understand how the app was working at all. I am testing my code in the browser, not in android studio, as the latter is too hard.
Antichess, Suicide and Giveaway differ in how the winner is determined when the active player can not make a move (stuck pawns), which can result in slight strategic differences. In the responses from the (apparently web only) stockfish I saw only the `UCI_variant`s of `giveaway`, `atomic`, `crazyhouse`, `horde`, `kingofthehill`, `racingkings` and `threecheck`. This also means that passing `3check` for the web platform is indeed broken.
To separate the issues, I will pull this logic out of this PR and put into another.

shabbychef:
So again, in the replies from the (web only) stockfish, it expects `threecheck`:
image

src/stockfish.ts
shabbychef: OK, peeled that out and put into PR #1833.

src/ui/ai/AiRound.ts
jas14: Hmm. There seem to be two things going on here: the local management of the board state via Chessground, and the replay engine (https://github.com/ornicar/scalachess). Calling `addMove` sends the promo move to the replay worker, which responds with the game situation information (including FEN), which eventually gets passed to `this.chessground.set` via the `onReplayAdded` -> `apply` callbacks.

So I think you're right - in practice, you don't need to specifically call promote, as it's all handled by the replay engine and callbacks. I suppose we still have to call `chessground.apiMove` for animation purposes? although `chessground.set` also does that.

src/ui/ai/AiRound.ts
jas14: It's in fact not obvious to me 🙂  Happy to help.

veloce:
`stockfish.js` is indeed old and does not match the new stockfish. It's only used for dev purposes, but I realize it can be confusing. 
Last release is from 2018: https://github.com/niklasf/stockfish.js/ so we have to deal with it.

veloce:
Thank you for this. Note that I had to revert the stockfish.js update @jas14 , it was breaking the offline computer for an unknown reason. It's actually not important to have the last version, because it's still old, and web stockfish is only used during web development.

PR #1833, opened 2021-09-01
Variant mangling

shabbychef:
It seems that the web stockfish understands different variant strings than the mobile stockfish-capacitor? In my testing within the browser, it was necessary to pass different UCI_variant strings.

### DIFF
--- a/src/stockfish.ts
+++ b/src/stockfish.ts
@@ -63,8 +63,10 @@ export class StockfishPlugin {
 
   public setVariant(): Promise<void> {
     if (this.isVariant()) {
-      if (this.variant === 'threeCheck')
+			if (Capacitor.platform !== 'web' && this.variant === 'threeCheck')
         return this.setOption('UCI_Variant', '3check')
+			if (Capacitor.platform === 'web' && this.variant === 'antichess')
+        return this.setOption('UCI_Variant', 'giveaway')
       else
         return this.setOption('UCI_Variant', this.variant.toLowerCase())
     } else {
--- a/src/stockfish.ts
+++ b/src/stockfish.ts
@@ -63,9 +63,9 @@ export class StockfishPlugin {
 
   public setVariant(): Promise<void> {
     if (this.isVariant()) {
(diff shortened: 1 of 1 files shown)


jas14:
Ah I see - we must not have updated stockfish.js in a long time! [Here](https://github.com/niklasf/stockfish.js/commit/965b0854874535719429ec24e7410effd296c2fb) is the commit that incorporates the upstream rename from "threecheck" to "3check." That was on April 24, 2017. 00ada4bd0a6eaa0a0403299ceb65521f4be5acad in this repository is when we last bundled stockfish.js - April 2, 2017. 

Let's just update stockfish.js to the [latest release](https://github.com/niklasf/stockfish.js/releases/tag/ddugovic-250718) - or to play it safe, the [release immediately after the threecheck rename](https://github.com/niklasf/stockfish.js/releases/tag/ddugovic-250417). That way, no special-casing will be required here 🙂

shabbychef:
I'm not sure I understand any of that. Nonetheless, in the commit you link to, I see suicide, antichess and losers chess (which I  think is giveaway).

jas14:
The Android and iOS app use a native [multivariant Stockfish fork](https://github.com/ddugovic/Stockfish). For browser testing, the app uses a "web plugin" called [stockfish.js](https://github.com/niklasf/stockfish.js), which is really just that multivariant Stockfish fork compiled into Javascript. We copy-paste that compiled Javascript file [into this repository](https://github.com/veloce/lichobile/blob/master/stockfish.js) At some point, the native plugins got ahead of the web plugin, which is why the web version expects `threecheck` but the native versions expect `3check`.

All this to say that instead of special-casing the three-check UCI for the web, we should just update the web plugin, stockfish.js, to the latest version - then it will accept `3check` as the native plugins already do. You can do this by replacing the `/stockfish.js` file in this repository with the (non-WASM) latest release of stockfish.js, found here: https://github.com/niklasf/stockfish.js/releases/tag/ddugovic-250718

shabbychef:
Thanks, I think I understand that.

shabbychef:
yay. thanks!

PR #1886, opened 2021-10-10
add Blindfold Chess mode

WnP:
A simple implementation of the blindfold chess mode available on the website.

It use css `display: none` on piece tags in order to hide pieces.

When enable pieces are hidden only during the game, which means that analysis, editor, TV or any watched game are not impacted.

Let me know your thoughts.


veloce:
It does perfectly the job, thank you.

PR #1943, opened 2021-12-04
Fix anonymous no-clock analysis screen shift

### LINKED ISSUE #1874: [Minor bug] Board jumps in analysis 'when not from the game' move is played because clock disappears
**Describe the bug**
The board jumps in analysis 'when not from the game' move is played because clock disappears and the position of the board is changed. It's annoying and never should happen (never happens on web-site of course).
Reproducible 100% at least using accont-less mode (not logged in).

**To Reproduce**
1. Play a game (at least using accont-less mode, not logged in)
2. The game is finished, go to analysis.
3. Make any move that was not played in the game.
4. The clock dissappers and the board changes it's position right after piece move.

**Expected behavior**
The board stays on the same place.

 - Device: Android
 - OS: Android 8.1
 - Version: 7.12.0

**Additional context**
This is regression in 7.12.0, caused by fixing this: #1759.

jas14:
Closes #1874. It's possible to be looking at an anonymous game with clock info for some nodes but with no clock info for the current node. Setting the player name to `&nbsp;` will prevent the screen from shifting between nodes with and without clock info for those games.

This code affects display for all analysis situations, so I've tried to capture all possible scenarios below. The two not listed that I can think of are:
- a non-synthetic anonymous game with no time control, which I don't believe it's possible to create (just tried to do so and was told "you need an account to do that")
- a study with clock info but no player info (don't know if it's possible to have a study with clock info?)

| Situation | Picture | Notes |
| ---- | ---- | ---- |
| Non-synthetic anonymous game | https://user-images.githubusercontent.com/569991/144712139-0e5d6636-3cdb-4e91-bfd6-d0bef7266b52.mov | **No screen jumps when move node has no clock info** - this is the new behavior |
| Study without player info |  | Player bars not included, as before |
| Study with player info |  | Player bars included, as before |
| Synthetic analysis (no player info) |  | Player bars not included, as before |
| Non-synthetic non-anonymous game | https://user-images.githubusercontent.com/569991/144712244-e4540efc-d1e2-4e60-b491-dd83742d716e.mov | No screen jumps, as before

### DIFF
--- a/src/ui/analyse/view/boardView.ts
+++ b/src/ui/analyse/view/boardView.ts
@@ -29,6 +29,12 @@ export function playerBar(ctrl: AnalyseCtrl, color: Color): Mithril.Child {
   const pName = ctrl.playerName(color)
   const isAnonymous = pName === 'Anonymous'
 
+  // Assumes studies without player info do not have clock info
+  if ((ctrl.study && isAnonymous) || (!ctrl.study && ctrl.synthetic)) {
+    return null
+  }
+
+
   const study = ctrl.study && ctrl.study.data
   let title, elo, result: string | undefined
   if (study) {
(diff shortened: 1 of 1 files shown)


veloce:
Thanks!

baack:
Thanks a lot!
Looking forward to get working analysis back in the next release, hope it won't take long.

Issue #1480, opened 2021-01-03
Puzzle v2


triplez73:
Does lichess plan to add puzzle v2 in the app ?
I means to offer the possibility to the user to choose which theme he wants just like the web version.

PS : does lichess plan to add a proper API for puzzle ?
Currently the api is only for getting puzzle activity ..

link2xt:
> I means to offer the possibility to the user to choose which theme he wants just like the web version.

Related issue about puzzle selection: #1499

PR #1958, opened 2021-12-23
Draft: Add option to change puzzle difficulty 

wbarnha:
Something that's been bothering me is that I have the option to modify the difficulty of puzzles on my Lichess browser application, but I can't do the same on mobile.

The code is currently pretty messy and not completely functional yet. Would there be any interest in this feature?


jas14:
I and many others have a lot of interest in [puzzles V2](https://github.com/veloce/lichobile/issues/1480) generally :) which includes:

- difficulty options
- theme selection
- [dashboard](https://lichess.org/training/dashboard/30/dashboard)
- [history](https://lichess.org/training/history)
- a lot of UI work, much of which we can probably take from the website

We might want to wrap this work up into that - up to veloce!

wbarnha:
I definitely think Puzzles V2 is a good thing to work towards, but it seems there's going to be some refactoring in the future to achieve that.

wbarnha:
I support the idea of moving forward with puzzles V2, but from what I currently see, it'll require a substantial change in how lichobile interfaces with the server.

wbarnha:
I saw the PR in https://github.com/lichess-org/lila/pull/10118. I like how there's more functionality to get puzzles batches by theme. But I'd still really like to see more added on for difficulty, so progress will need to wait.

wbarnha:
I'm closing this PR because 1) it's getting messy, and 2) there's some additional work that needs to be done in localization before I'd enable support for themes. The backend for lila still needs some work some to support difficulty, but we can take measures to support puzzle themes in the meantime.

PR #1980, opened 2022-01-13
Remove unneeded Android theme parameters

kevinslashslash:
These aren't doing anything productive, as they're quickly overridden when
`BridgeActivity` resets the theme. Additionally `windowIsTranslucent` may trigger
this Android issue https://issuetracker.google.com/issues/194427611 , when
using 3rd party launchers that support the return to home animation on
Android 11+, the Lichess icon may get stuck on the screen.

Tested on Android 11+ with and without GestureNavContract and with and without dark mode.
Also tested on Android 8.0 to verify this doesn't introduce splash screen issues on older Android versions.
I was unable to find out why these parameters were originally added.

### DIFF
--- a/android/app/src/main/res/values/styles.xml
+++ b/android/app/src/main/res/values/styles.xml
@@ -18,10 +18,5 @@
 
     <style name="AppTheme.NoActionBarLaunch" parent="AppTheme.NoActionBar">
         <item name="android:background">@drawable/launch_splash</item>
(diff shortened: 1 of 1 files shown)


veloce:
It's not anymore in the [capacitor template](https://github.com/ionic-team/capacitor/blob/main/android-template/app/src/main/res/values/styles.xml) indeed.

PR #1984, opened 2022-01-14
Fix tournament header padding on notched devices

### LINKED ISSUE #1979: Tournament profiles padding
I guess this one is similar to #1970. To reproduce open any tournament and click on any participant.

photo_2022-01-12_17-36-27

jas14:
Closes #1979.

Since we're using [the default `box-sizing: content-box`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing), we don't need to adjust the height to account for the padding, which is already set to `env(safe-area-inset-top)`.

## Before

Simulator Screen Shot - iPhone 11 - 2022-01-14 at 09 22 16

## After

Simulator Screen Shot - iPhone 11 - 2022-01-14 at 09 40 28

### DIFF
--- a/ios/App/Podfile.lock
+++ b/ios/App/Podfile.lock
@@ -41,26 +41,26 @@ PODS:
     - Capacitor
   - CapacitorToast (1.0.6):
     - Capacitor
-  - Firebase/CoreOnly (8.9.1):
-    - FirebaseCore (= 8.9.1)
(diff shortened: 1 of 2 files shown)


veloce (commented):
It would be better to have only the style change in this PR and no `Podfile` change. Thanks!

Issue #1993, opened 2022-01-17
Text readability in studies and move history.


2i3r:
**Is your feature request related to a problem? Please describe.**
Reading through studies (comments)/game moves, is hard, as readability is pretty low. font size is monotone, new line on most of studies doesn't work,  and there is not much text format. 

**Describe the solution you'd like**
Adding support for markdown styling, font size settings, and replacing incompatible new line/return carriage, should resolve problem.

**Describe alternatives you've considered**
font size settings alone might help a lot. 

**Additional context**

![edited2](https://user-images.githubusercontent.com/16647235/149787299-2b666129-5769-4945-9ecf-8035918f7c14.jpg)

veloce:
Closing as new lines have been implemented.

PR #1995, opened 2022-01-18
Render newlines as HTML line breaks

jas14:
Part of #1993.


src/ui/analyse/view/treeView.ts
```diff
@@ -58,28 +58,14 @@ export default function renderTree(ctrl: AnalyseCtrl): Mithril.Children {
   ])
 }
 
-function renderInlineCommentsOf(ctx: Ctx, node: Tree.Node, rich?: boolean): MaybeVNode[] {
```
jas14, line 61: `rich` was `true` at all callsites.

Issue #2000, opened 2022-01-21
Android app game goes rogue, time/ move chaos.


surfstian:
**Describe the bug**
Like typical network issue, just much worse.

**To Reproduce**
1. Start game
2. Don't be able to move and make clock go crazy
3. Loose game

**Expected behavior**
The opposite.

**Screenshots**
See video

**Smartphone (please complete the following information):**
 - Device: Galaxy S10
 - OS: Android 11
 - Browser: App
 - Version v7.14.1

**Additional context**
This also happens randomly at a lesser scale.  Even with a network issue it should not go nuts like this? 

There could at least be som sort of info and /or report on the issues for the game.

https://user-images.githubusercontent.com/19313191/150551160-1f781e3c-a43c-4f80-9322-b81575c4d2c1.mp4

surfstian:
https://user-images.githubusercontent.com/19313191/150551417-5a6e4f9e-1591-4dfc-b55a-2b125a12e003.mp4

jas14:
It seems your connection is very poor?

surfstian:
No matter what the connection does the game should not behave in this manner? Anyway, the dots appear about mid game, the problems are present for the whole game. It should be clear immediately that the app is not able to sync with server?

Also as time goes down the bugging seem to increase and I keep getting that f#%@%@ bouble animation that locks up an ruins the game.

Sometimes when I make a move the clock keeps going, I mean - why is the app not even able to register my move, give a fault indication or what ever, it knows I made the move and must know the move does not go through because the @#^^$$& clock keeps going. 

:)

Just trying to help here.

jas14:
I understand - it can be very frustrating when it looks like it's still your turn and your clock is still counting down after you've moved!

The way the website handles being totally offline is to stop both clocks until it comes back online. We could certainly implement that.

The website also deals with the case of a very slow connection by stopping both clocks as soon as the user makes a move, resuming the clocks only when it receives an acknowledgement of the opponent's move (which also contains new clock time information). We could also implement that.

jas14:
Probable duplicate of #1916.

bonstio:
Loads of evidence of this happening on https://github.com/lichess-org/lichobile/issues/1916 where I've captured videos of the behaviour.

[...]

PR #2014, opened 2022-02-06
Stop clocks until move readback is received

jas14:
Addresses #2000 and probably some other reports.

Currently, the app shows the current player's clock as active until it receives the move "readback" from the server. On high-latency connections, this can be confusing, as it may appear that the app hasn't registered or sent the move.

The solution implemented on the desktop site in https://github.com/lichess-org/lila/commit/0061c263b87953c1d8ec1368ed25cf837b7d6a33 / https://github.com/lichess-org/lila/pull/3300 was to stop the clocks when sending a move, which is what this PR mirrors.

Tested locally with network conditioning - will attach a screen cap shortly.

See:
- [lila round control `actualSendMove`](https://github.com/lichess-org/lila/blob/cd6e215c691141c76dcd3ee946b39a60ebca02d4/ui/round/src/ctrl.ts#L326)
- [lila clock view](https://github.com/lichess-org/lila/blob/bce9fd8f0ebe8d9766bad1ebb00051185cc33ac0/ui/round/src/clock/clockView.ts#L14)

### DIFF
--- a/src/ui/shared/round/OnlineRound.ts
+++ b/src/ui/shared/round/OnlineRound.ts
@@ -702,6 +702,12 @@ export default class OnlineRound implements OnlineRoundInterface {
       blur
     }
 
+    // Pauses both clocks. When the server echoes back this move, the clocks will resume.
+    // This prevents a confusing user experience where a laggy network makes it appear that
+    // it's still the user's turn; see #2000.
+    this.clock?.stopClock()
+    redraw()
+
     if (isMoveRequest(moveOrDropReq)) {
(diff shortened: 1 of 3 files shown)


jas14:
https://user-images.githubusercontent.com/569991/152689709-c53cbe27-acd1-42ea-818c-84ca2dae673d.mov

veloce:
I didn't know the network throttling was applied to websockets. I guess it's been fixed in a recent version, that's nice!
I'll do some other tests as it's a sensitive change. Thanks for this!

PR #2062, opened 2022-03-15
Retry fixing #1869 (bad display on square screen)

### LINKED ISSUE #1869: display on square screen
**Describe the bug**
I got a new phone with a square screen : a Unihertz Titan (android 10)

Here is the display I get.
[Screenshot_20210929-221243](https://user-images.githubusercontent.com/91645413/135375205-7169c610-6650-40ad-9bfe-e5f51d5d3c27.png)

Lichess app is up to date.
I didnt find any way to reduce the size of the board in order to access the different shortcuts (like share, analyse and so on).
, whatever the menu is set on left or right in options, it still look like compressed.

Is this a display bug or did I miss something? Thanks

alexpdp7:
Sorry @veloce , I didn't give you the proper data to fix #1869. Although the screen size is reportedly 716x720, it seems that the viewport size is just 665x716 (as reported by Chrome when doing remote debugging)

After applying this change, this is how the screen looks:

image

I have no idea about media queries, so I'm not exactly sure why you pinned the aspect-ratio between 716/720 and 720/716, but I have preserved that. Happy to make more tests if needed, now that I figured out how to build/debug.

### DIFF
--- a/src/styl/board-content.styl
+++ b/src/styl/board-content.styl
@@ -51,7 +51,7 @@
     width var(--board-side-land-tablet)
     margin auto
     margin-left 3vh
-  @media only screen and (min-aspect-ratio: 716/720) and (max-aspect-ratio: 720/716)
+  @media only screen and (min-aspect-ratio: 665/716) and (max-aspect-ratio: 716/665)
     height 60vw
     width 60vw
     margin auto


jas14:
> However, neither that nor the APK I built there is unable to connect to lichess.org, so I cannot log in to my account and test this more heavily.

**Disclaimer: testing in production is not generally recommended.**

That being said, the default appconfig.dev.json points to lichess.dev. You can create your own e.g. `appconfig.prod.json` to point to the Lichess production servers (IIRC `https://lichess.org` for API, `wss://socket0.lichess.org` for websocket), then run `APP_CONFIG=prod npm run build` to inject the prod configuration into the compiled webapp and mobile apps.

alexpdp7:
Oh, I should have figured that out :(

Honestly, even though I can see the docs on running your own Lichess instance, I just thought it might be too much trouble for casual contributions. Also, I was thinking of just building me an APK until this is merged and released, so I can play :)

*IF* a separate PR to extend the docs explaining how to point to lichess.org is welcome, I can do that. (With a big disclaimer on top). But I can see that some might prefer to leave that undocumented, so up to you.

In the meantime, if the PR needs any further work to be reviewed, happy to do that too.

jas14:
Yeah, unless you're hosting your own lichess instance on a server easily accessible from your phone, it's not worth it. I've found it useful for testing via the webapp or emulator on your local machine or emulator, but not so much for a real device.

> IF a separate PR to extend the docs explaining how to point to lichess.org is welcome, I can do that

I'll have to defer to the maintainer of this repo (veloce) and/or ornicar on that.

I'll also defer to veloce on this PR itself - I'm not sure of all the implications of widening the "square" aspect ratio, and he is the final authority on this repo anyway. 🙂

veloce:
@alexpdp7 does it mean that my last release didn't change anything for your screen?

If I used a min and max in the media query, it was to handle the screen rotation: since it's not perfectly square we need to deal with that.

alexpdp7:
Yeah @veloce ; the media query didn't apply to my screen. I think the root of the issue is that while the screen size is the one I told you (716x720), the status bar and bottom gesture thing reduce the viewport to less than that, so the media query does not apply, as it was very "strict".

Your rule applied to aspect ratios from 0.9944 to 1.0056 (716/720 to 720/716). 665/716 is 0.92 actually, so it lies outside that.

I suspected the range was due to portrait/landscape. For this particular smartphone, it never makes sense to use landscape (it's got a keyboard south of the screen, so if you do landscape, you have a sideways keyboard to the right... and as the screen is [nearly] square, you don't gain much).

I can do some further testing and tell you the exact viewport on landscape (it's probably not 716x665, the status bar and bottom gesture thing will likely eat a different amount of space).

veloce:
The screenshot you sent appears as landscape: in the app landscape mode, the board is on the left and controls on the right. In portrait the board would be on top with controls below. To me your changes are fine, I'll merge as is. Thanks!

Issue #1918, opened 2021-11-18
Use nicer taptic feedback on iOS instead of generic “vibrate”


ulope:
**Is your feature request related to a problem? Please describe.**
When vibration is enabled a generic (and relatively long) “vibrate” is used.

**Describe the solution you'd like**
All current iPhones have a Taptic Engine that allows much nicer / richer haptic feedback.
For example for an opponents move a simple tap could be enough while a check might be a stronger double tap.

jas14:
@ulope thanks for the feedback. If you're interested, this should be "fairly straightforward" (famous last words) to implement! Capacitor already has a [Haptics plugin](https://capacitorjs.com/docs/apis/haptics), which is what we use. We only use the `vibrate` method right now, with the default 300ms duration. You could certainly vary the duration based on move type, maybe use an "impact"-type vibration, or even do more complex "double-tap" sort of haptics as you've mentioned. It would all be contained in the TypeScript code, meaning you could write once for both platforms. 😃

ulope:
Ok I might have a look at that, although I don't have any experience with capacitor.
The docs aren't very exhaustive it seems. They mention that on platforms that lack enhanced taptics support the API calls silently do nothing. 
Do you know a way to query for supported features?

jas14:
I don't know of a way to query for supported features, but fortunately I don't think that's necessary in this case. If a device doesn't support vibration at all, the user would presumably know and would not be surprised if there were no vibration when the setting is enabled. Furthermore, from a cursory look at the `@capacitor/haptics` source code for [Android](https://github.com/ionic-team/capacitor-plugins/blob/3beb75a9bfc18d0586f2c9df1b584250e95e0237/haptics/android/src/main/java/com/capacitorjs/plugins/haptics/Haptics.java) and [iOS](https://github.com/ionic-team/capacitor-plugins/blob/3beb75a9bfc18d0586f2c9df1b584250e95e0237/haptics/ios/Plugin/HapticsPlugin.swift), it appears all of the haptics interfaces are implemented on both platforms, so you don't have to worry that certain calls will work on one platform but not the other.

So the TL;DR good news is that you just have to focus on the JS-side calls.

PR #2079, opened 2022-04-04
Make game events emit haptic effects

ulope:
Instead of a generic 'vibrate' game events now emit different haptic effects.

There are different effects for:
- Regular move
- Capture
- Check
- Draw
- Win
- Lose

Fixes: #1918

Caveat: 
I have very little idea what I'm doing here. Haven't used capacitor before and it's been quite a while since I touched mobile dev.
Seems to work as expected though, both in "play against computer" and online play.


veloce:
It looks good! Going to test this now.

veloce:
Sorry there will be some delay on this one as I have currently an issue where Xcode doesn't let me run the app on my device. I think I need to update it but I have no stable connection at the moment to download the 12G.

PR #2091, opened 2022-04-17
Implement nice haptics for AI games

lapfelix:
Addresses https://github.com/lichess-org/lichobile/issues/2090

In https://github.com/lichess-org/lichobile/pull/2079 the nicer haptics are used only on online games. This PR does the same for computer games.

I moved the vibration code from `onMove()` to `apply(sit:)` so we can check wether the last move is a check (and double tap when it's the case).

### DIFF
--- a/src/ui/ai/AiRound.ts
+++ b/src/ui/ai/AiRound.ts
@@ -286,7 +286,6 @@ export default class AiRound implements AiRoundInterface, PromotingInterface {
     } else {
       sound.move()
     }
-    vibrate.quick()
   }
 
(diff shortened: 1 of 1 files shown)


veloce:
Nice! If we do it in `AiRound.ts` it should be done in `OTBRound.ts` as well. Thanks.

veloce:
Closing with https://github.com/lichess-org/lichobile/commit/61b4f31dc3e986780db606e5129eb68f748496ac

It's not possible to use `apply` unfortunately because it would trigger a tap on each move during replay.

PR #2093, opened 2022-04-19
Add puzzle themes to localizations

wbarnha:
I closed #1958 since there's some working needing to be done on the back-end before mobile users can request puzzle difficulties. This is a basic step forward to including localization for puzzle themes since https://github.com/lichess-org/lila/pull/10118 indicates that there's support for mobile clients requesting themes. Or should we hold off on merging this until changes include puzzle theme requests? It seems unnecessary to include this code until we're using it, but I'll let the maintainers decide.


wbarnha:
I also noticed the en-US and en-GB localizations aren't identical. Should we change that?

jas14:
@wbarnha thanks for the PR! I don't think we should add theme localizations before they're used anywhere.

Anyway, for when we _do_ start using them, the process is:

1. Add the relevant [translation module](https://github.com/lichess-org/lila/tree/c48da4e2cd22837a9f6e8b7eb4a59924967aec3f/translation/dest) name (in this case, [`puzzleTheme`](https://github.com/lichess-org/lila/tree/c48da4e2cd22837a9f6e8b7eb4a59924967aec3f/translation/dest/puzzleTheme)) to [the list in `scripts/i18n-update.mjs`](https://github.com/lichess-org/lichobile/blob/b6df2c43d7f8ddf4682aeb58cc3ce93882e3da73/scripts/i18n-update.mjs#L18)
2. Run `npm run i18n-update`

This will download and unpack the relevant translations for all languages from the lila project. Sorry this isn't well-documented, would have saved you some work!

jas14:
> I also noticed the en-US and en-GB localizations aren't identical. Should we change that?

No, there are a handful of spelling differences between the two. Check out `toReportSomeoneForCheatingOrBadBehavior` for example: "behavior" vs "behaviour."

While we _internationalize_ (most of) the code in this repo, we don't manually manage translations. That's done by contributors (anyone who wants!) over on [Crowdin](https://crowdin.com/project/lichess). If you think a translation is incorrect, or in this case that there's an incorrect discrepancy between en-US and en-GB, feel free to contribute there!

PR #2144, opened 2022-07-08
Correct the route for study chapters

djpohly:
Chapter links from the website don't have `/chapter/` in the URL. The app exports them correctly but does not recognize them.


veloce:
Thanks

PR #2151, opened 2022-07-21
Allow building without firebase-messaging.

gilbsgilbs:
**Disclaimer:** This is a proposal more than anything else. I think the added maintenance burden is rather minimal, but if you think otherwise that's OK. I can as well maintain my own fork with these changes. Feel free to close. No pressure, no nothing 🙂 .

Also, as I said in [this comment](https://github.com/lichess-org/lichobile/issues/1121#issuecomment-1190883854), I think the push stub is pretty equivalent to having the "Notifications" checkbox disabled in the current Play Store version which sounds quite harmless. But I'm still not 100% sure `pushNotificationActionPerformed` hook cannot be invoked when not registered.

Also, I was able to build the debug version without issue, but the release version won't build on my machine at the moment, even on master branch with the following error:

```
> Task :app:lintVitalRelease FAILED
/home/nleroux/projects/lichobile/android/app/src/main/AndroidManifest.xml:27: Error: MainActivity must extend android.app.Activity [Instantiatable]
            android:name="org.lichess.mobileapp.MainActivity"
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/nleroux/projects/lichobile/android/app/src/main/AndroidManifest.xml:84: Error: FileProvider must extend android.content.ContentProvider [Instantiatable]
            android:name="androidx.core.content.FileProvider"
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

   Explanation for issues of type "Instantiatable":
   Activities, services, broadcast receivers etc. registered in the manifest
   file (or for custom views, in a layout file) must be "instantiatable" by
   the system, which means that the class must be public, it must have an
   empty public constructor, and if it's an inner class, it must be a static
   inner class.

2 errors, 0 warnings
Lint found fatal errors while assembling a release target.
```

This should probably be addressed in another PR, otherwise the F-Droid recipe might be failing.

~Also, I noticed that `libstockfish.so` was only present in the APK for the arm64 architecture (even on the Play Store release). Is that normal or is this a bug? Does stockfish work correctly on non-arm64 systems?~ After checking, it uses "stockfish-variant" (without NNUE) as a fallback. I opened https://github.com/veloce/capacitor-stockfish/pull/8 to build NNUE on other platforms.

---

This commit provides a way to build lichobile for Android without
support for push notifications, dropping the dependency on Firebase.
This is a required step towards F-Droid inclusion.

The main idea is to expose a `patch-nonfree` script that will:

- replace the push notification implementation with a stub
- remove any direct or indirect dependency to firebase (there was an
  indirect one in the @capacitor/push-notifications plugin, and a direct
  one in the app/build.gradle)
- change the applicationId so that it doesn't gets confused with the
  non-free one

The implementation is not as elegant as I'd have liked because I haven't
found a magic way to enable or disable dependencies for specific product
flavors with capacitor ([as the Android gradle plugin allows](
https://developer.android.com/studio/build/build-variants#product-flavors)).
I'm far from a capacitor expert though, so there may be better ways.

As I had to patch the JS sources anyways, I chose to not rely on
Android's product flavors at all to prevent any confusion that would
lead to build a non-free version as a free APK and vice-versa. I think
it is pretty straightforward, not too intrusive, and that it should be
manageable maintenance-wise.

The remaining changes in the JS code are needed to remove any mention of
"notifications" in the UI when using the push stub. We don't expect
notifications to work on the free version.

Fixes https://github.com/lichess-org/lichobile/issues/1121

(well, technically this doesn't "remove" the dependency to Firebase as
the main build flavor still relies on it, but I think getting rid of
that is lengthier project as lila itself depends on it)

Issues related to F-Droid inclusion:

- https://github.com/lichess-org/lichobile/issues/278
- https://github.com/lichess-org/lichobile/issues/409
- https://github.com/lichess-org/lichobile/issues/927
- https://github.com/lichess-org/lichobile/issues/1331

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -6,6 +6,7 @@ on:
   push:
     paths:
       - '.github/workflows/android.yml'
+      - '.github/actions/build-lichobile-apk.yml'
       - 'android/**'
     branches-ignore:
       - 'l10n_master'
@@ -14,54 +15,22 @@ jobs:
   apk:
     name: Generate debug APK
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Node.js
-        uses: actions/setup-node@v2
+      - uses: ./.github/actions/build-lichobile-apk
+
+  freeApk:
+    name: Generate free debug APK
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/build-lichobile-apk
         with:
-          node-version: '16'
-          cache: 'npm'
-      - name: Install Dependencies
-        run: npm ci
-      - name: Build web
-        run: npm run build
-      - name: Setup JDK
-        uses: actions/setup-java@v2
-        with:
-          distribution: 'temurin'
-          java-version: '17'
-          cache: 'gradle'
-      - name: Sync android capacitor
-        run: npx cap update android
-      - name: Prepare build
-        working-directory: ./node_modules/capacitor-stockfish/stockfish/src
-        run: make net
-      - name: Build APKs
-        working-directory: ./android
-        run: |
-          ./gradlew assembleDebug
-      - uses: actions/upload-artifact@v2
-        with:
-          name: debugApk
-          path: |
-            ./android/app/build/outputs/apk/debug/
-      - name: Patch nonfree
-        env:
-          PATCH_NONFREE_FORCE: "1"
-        run: npm run patch-nonfree
-      - name: Sync android capacitor
-        run: npx cap update android
-      - name: Build Free APKs
-        working-directory: ./android
-        run: ./gradlew assembleDebug
+          artifactName: "debugFreeApk"
+          patchNonFree: "true"
       - name: Check non-free dependencies
         working-directory: ./android/app/build/outputs/apk/debug/
         run: |
           dexdump=$(find $ANDROID_HOME/build-tools/ -name 'dexdump' | sort -r | head -1)
           ! $dexdump app-debug.apk | grep 'Class descriptor' | grep com\/google\/firebase
           ! $dexdump app-debug.apk | grep 'Class descriptor' | grep com\/google\/android\/gms
-      - uses: actions/upload-artifact@v2
-        with:
-          name: freeDebugApk
-          path: ./android/app/build/outputs/apk/debug/
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -46,3 +46,22 @@ jobs:
           name: debugApk
           path: |
             ./android/app/build/outputs/apk/debug/
+      - name: Patch nonfree
+        env:
+          PATCH_NONFREE_FORCE: "1"
+        run: npm run patch-nonfree
+      - name: Sync android capacitor
+        run: npx cap update android
+      - name: Build Free APKs
+        working-directory: ./android
+        run: ./gradlew assembleDebug
+      - name: Check non-free dependencies
+        working-directory: ./android/app/build/outputs/apk/debug/
+        run: |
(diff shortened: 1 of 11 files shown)


EvanCarroll:
Not a useful comment, but thanks ton for your work here @gilbsgilbs. I look forward to using f-droid on lichess. And to say this will be an asset to f-droid is an understatement. Wouldn't shock me at all if this is the largest game on the platform.

veloce:
Thanks for this @gilbsgilbs , as I know several people already asked for that. I cannot tell you yet if I will merge it or not, I'll need to have a closer look at it. If this patch doesn't require extra maintenance/work for me I'll happily merge it.

.github/workflows/android.yml
```diff
@@ -46,3 +46,22 @@ jobs:
           name: debugApk
           path: |
             ./android/app/build/outputs/apk/debug/
+      - name: Patch nonfree
```
veloce, line 49: Can you please separate this from the normal android job? Both free and non-free builds should be separated.
The new job would trigger the same way as the normal one.

gilbsgilbs: Done in 762eff2344b9b5db6ca7791729aaa5160564a2d3

jas14:
I'm going to investigate how much of this can be done with Gradle build variants/flavors.

gilbsgilbs:
@jas14 Have you figured out a way, or do you have some specific idea to achieve this?

For the records and to give a bit more context: I initially wanted to implement this through build flavors, but I eventually renounced. Firstly, it would require to get the build flavor from the TS code which already sounds tedious yet doable (at very least by having the flavor set a [buildConfigField](https://developer.android.com/reference/tools/gradle-api/4.2/com/android/build/api/variant/BuildConfigField) and retrieving it from a native module). But then, the project would still include `@capacitor/push-notifications` gradle files, which depend on firebase. You can probably easily use some dirty gradle trick to prevent firebase from being included, but then `@capacitor/push-notifications` would fail to build, so something needs to be done. I had some candidates workarounds in mind:
1. "uninclude" `@capacitor/push-notifications` and patch any code that'd depend on it or mock it entirely at build time.
1. create a `@capacitor/push-notifications` wrapper plugin that could switch to a mock implementation if given some flag.
1. create a `push-notification-mock` plugin and switch between the implementations somehow

Both workarounds sounded very cumbersome and "backwards" to me, if possible at all. Moreover, I assumed that no matter how I would be able to implement this in practice, it would always be way more complex, prone to chaotic breakages and to increased maintenance burden than just straightly patching out the dependency and do the cap sync manually as I did in this PR. Not only had it less chances of being merged upstream, I thought that a plain fork would 100% be preferable to such mess anyhow.

I even considered an hybrid approach (using a build flavor just for the app id suffix and the build flag) which I initially did, but that doesn't do anything for `@capacitor/push-notifications` that you still need to patch out which is already 95% of what this PR does, so that's just plain over-engineering. I concluded it was pointless to rely on build flavors at all if `gradlew assembleRelease` wasn't able to create the two flavors at once, which, as I shown, looks very complex to achieve in practice from my (poor) knowledge of capacitor.

If you have neat ideas, I'm happy to help and rework on that: I might very easily have missed the obvious with capacitor. However, I think maintenance-wise, what's implemented in this PR is a lot more preferable than any of the other ideas I had. And again, if you feel like it's too much maintenance burden, we can discuss on a fork and how to brand it.

jas14:
You're probably spot on - I wasn't sure if you'd tried it. I vaguely understand different code can be included in different flavors, but 1) it's true that this would introduce maintenance burden and 2) it may be quite difficult to use capacitor with flavors.

veloce:
I tested the android build and it's fine, so I'm merging.

PR #2160, opened 2022-08-05
Add setting to disable back button during game

dancek:
This is very useful on Android 12 without dedicated buttons, as swiping
from the edge of the screen generates a back button press event. Under
time pressure moving pieces on the edge of the board may easily cause
navigation out of the game and greatly affect the experience. I've lost
a dozen blitz games like this.

With the setting enabled, a back button press simply does nothing which
is strictly better. Even better would be to disable the swiping gestures
during game altogether, but I don't know of a way to do that.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -7,6 +7,7 @@ import isFunction from 'lodash-es/isFunction'
 
 import signals from './signals'
 import session from './session'
+import settings from './settings'
 import { serializeQueryParameters } from './utils'
 import redraw from './utils/redraw'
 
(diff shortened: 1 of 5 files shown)


dancek:
Anything I can do to get this PR merged?

mormubis:
I need this feature. It is really annoying when playing bullet games and I drag a piece from the border of the screen triggering the back button.

dancek:
@veloce can I do something to help get this PR merged?

veloce:
Thanks for the PR. The idea was good but I prefer not having another setting to deal with this. I have made another PR that fixes the issue more transparently: https://github.com/lichess-org/lichobile/pull/2257

PR #2191, opened 2022-09-12
Add fallback values for promo menu

eljamm:
Old Android WebView versions might not support [calc](https://caniuse.com/calc) or [math](https://caniuse.com/css-math-functions) functions properly, if at all, which makes the promo menu appear empty.

To solve this, either the user needs to update their WebView to the latest version or we can add fallback values as a catchall.

These are the tests I've done on my Android 8.1 device :

* WebView 70 + 1.7.16 release -> Empty Menu
* WebView 70 + 1.7.16 release (with fallbacks) -> Correct Menu
* WebView 104 + 1.7.16 release -> Correct Menu

I don't know if this fully fixes the empty menu issue, but for my device it appears to be the case.

### DIFF
--- a/src/styl/game.styl
+++ b/src/styl/game.styl
@@ -302,9 +302,11 @@ $material-piece-size = 16px
   column-gap 5px
   piece
     position relative
+    width 60px // fallback
     width calc(min(12.5%, var(--overlay-height)))
     // CSS hack: padding-bottom is calculated based on width of parent, so we can get a square div here
-    height 0
+    height 60px // fallback
+    height calc(min(0px, 60px))
     padding-bottom calc(min(12.5%, var(--overlay-height)))
 
 .tableHeader


src/styl/game.styl
```diff
@@ -302,9 +302,11 @@ $material-piece-size = 16px
   column-gap 5px
   piece
     position relative
+    width 60px // fallback
     width calc(min(12.5%, var(--overlay-height)))
     // CSS hack: padding-bottom is calculated based on width of parent, so we can get a square div here
```
veloce, line 307: I don't understand the `CSS hack` comment, but changing from `height 0` to `height 60px` is safe here?

eljamm: The CSS hack refers to making a [responsive square div](https://www.dwuser.com/education/content/creating-responsive-tiled-layout-with-pure-css/). Here, the actual height is calculated as
`padding-bottom + height` and we want the div to be square so we need the height to be 0.
You can play around with this concept in [jsfiddle](https://jsfiddle.net/x6jj3dar/).

Old versions of WebView will ignore the lines that include calc and min, so their width and height will be 60px, as per the behaviour before #2111.

```styl
# Old Versions
width 60px
height 60px
```

New versions will overwrite the fallback properties for width and height, in this case they would become :

```styl
# New Versions
width calc(min(12.5%, var(--overlay-height)))
height calc(min(0px, 60px))
```

What's important here is that we need what's inside calc() for height to evaluate to 0px in the end, so `calc(min(0px, 60px))` will always be `0px`. For example we can also do `height calc(min(0px, 0px))`.

veloce: Got it. 
I doubt the problem comes from old webview versions, because the oldest version supported by the app is 69, and according to MDN `calc` is supported from android chrome version 28.
I guess it's more like a bug on some specific versions of the webview.

I'll do one more test on my device to confirm it's working and then merge it. Note that I'll be AFK for 2 weeks so this will have to wait for the release.

eljamm: I see, I didn't research well enough on the specific versions that don't support it so it's probably as you said, but it would be nice to have a catchall for them nonetheless.

Hope this has been useful in fixing the issue or at least in shedding more light on the situation.

Pavelabramovich:
I had the same problem but this method helped me

Issue #2083, opened 2022-04-08
allow user TV open in app


valvin1:
When I want to watch a user TV by clicking on a link it is not opening in the app (Android on my case but it should be the same with other OS).

So if I click on https://lichess.org/@/valvin1/tv browser offers me only browser apps to view this page althought it is implemented in the app.

I add a quick look on Android side and it looks it requires adding url pattern here: https://github.com/lichess-org/lichobile/blob/master/android/app/src/main/AndroidManifest.xml#L40

Wondering if it could be an easy quick win :)

tleadon:
Hello! I'd like to pick up this issue, but its strange. 
"/tv" is [already included in the URL patterns in the manifest file.](https://github.com/lichess-org/lichobile/blob/master/android/app/src/main/AndroidManifest.xml#L63) Where else would one look to address this issue?

jas14:
Excellent! The path prefix is relative to lichess.org - so lichess.org/tv will be picked up by the app, whereas lichess.org/@/<username>/tv won't. (Yet!)

tleadon:
Ohhhhh! That makes sense! I'll get right on that!

PR #2197, opened 2022-09-22
Issue #2083 added a /@ datatag for app launching improvements 

tleadon:
#2083 
Added 
<data
         android:host="lichess.org" android:scheme="https"
         android:pathPrefix="/@" />

to solve the issue of links like https://lichess.org/@/valvin1/tv not launching the mobile app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,10 @@
+> [!WARNING]
+> ## Deprecation notice
+>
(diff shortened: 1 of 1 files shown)


jas14:
Have you tested this? I think this will intercept all user profile, TV, and potentially even more routes. We need to make sure that if the app intercepts these routes, it will handle them correctly, i.e. open to the correct page.

Issue #2074, opened 2022-03-31
Piece promotion view distorted, misaligned due to large promotion images 


GregWeiher:
**Describe the bug**
Piece promotion view distorted, misaligned due to large promotion images, see screenshot

**To Reproduce**
1. Any Attempt to promote a pawn regardless of piece colour, results in issue

**Expected behavior**
Used to be formatted correctly, 

**Screenshots**
See Screenshot attached
![Screenshot_20220331-075304_lichess](https://user-images.githubusercontent.com/12046645/160993038-8ac68cf2-b6d4-4965-a9fe-8a0c6827c5d3.jpg)

**Smartphone (please complete the following information):**
 - Device: Samsung S21 FE
 - One UI Version: 4.0
 - Android Version: 12
 - Google Play System Update: 01 Feb 2022
 - Baseband Version G990EXXS1BVB5
 - Kernal Version: 5.4.129-23339874-abG990EXXS1BVB8 \# 2 Fri Feb 18 10:19:23 KST 2022
 - Build Number SP1A.210812.016.G990EXXS1BVB5

**Additional context**
Post update of device and then lichess App, the promotion view for both black and white have been misaligned/distorted

jas14:
@GregWeiher thanks for the detailed report, looks like people are having a similar problem on #2059. I'll leave this issue open since this is on an S21, but you might have some luck with the workarounds described over there.

GregWeiher:
@jas14 
apologies on the delayed response, issue seems to relate directly to the font size as per the pictures below:
Thus anything other than the smallest font size, will distort the lichess apps view. interestingly, I hadn't realised that the issue was not only with the piece promotion but with everything font related in the app, and it appears that everything is scaled according to the font size setting. curiously though the promotion issue above seems to persist even if the font sizes are changed. 

![Screenshot_20220515-121053_Settings](https://user-images.githubusercontent.com/12046645/168467738-fa98e2c2-61b2-48df-8556-98ad493defbf.jpg)
![Screenshot_20220515-121059_lich
![Screenshot_20220515-121117_Settings](https://user-images.githubusercontent.com/12046645/168467746-223befc7-a673-4966-b094-22b5f0f9538e.jpg)
ess](https://user-images.githubusercontent.com/12046645/168467743-a0f69c60-8abe-4843-b240-2b57b4947f9c.jpg)
[...]

PR #2225, opened 2022-10-25
Revert "Scale promo piece size relative to screen width"

### LINKED ISSUE #1793: Promotion menu empty
Reported here: https://lichess.org/forum/lichess-feedback/bug-unable-to-promote-pawn#1
> The field there i should be able to choose figure is empty. First time i found this bug a few months ago and it repeats every time im trying to promote my pawns. That's why i hate late game 😅
I've already tried to reinstall lichess app, but that didn't solve the problem.
Also, I've got no problems in puzzles — I'm able to promote pawn there, so i have no idea why this field is empty in normal games

Android 8.1.0
redmi 6
lichess app version 7.11.3 (latest)

Maybe fixed by https://github.com/veloce/lichobile/pull/1785 ?

jas14:
Reverts lichess-org/lichobile#2111. Reopens #2074, closes #1793.

### DIFF
--- a/src/styl/game.styl
+++ b/src/styl/game.styl
@@ -289,23 +289,20 @@ $material-piece-size = 16px
     background-image url(../../images/pieces/mono/K.svg)
 
 #promotion_choice
-  --overlay-height 100px
   border 1px solid black
   background-color rgba(255,255,255, 0.7)
   position absolute
-  top calc((100vh - var(--overlay-height)) / 2)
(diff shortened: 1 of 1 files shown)


veloce:
Let's wait to see if #2191 solves the issue. If not, we'll revert. What do you think?

jas14:
Sounds good, I would much rather fix forward than revert

PR #2226, opened 2022-10-25
Fixed vulnerabilities CVE-2022-22912: Plist can cause denial of service

theinternetofdefcon:
## Describe the bugs:
Prototype pollution vulnerability via `.parse()` in Plist allows attackers to cause a Denial of Service (DoS) and may lead to remote code execution. 

Prototype Pollution is a vulnerability affecting JavaScript. Prototype Pollution refers to the ability to inject properties into existing JavaScript language construct prototypes, such as objects. JavaScript allows all Object attributes to be altered, including their magical attributes such as __proto__, constructor and prototype. An attacker manipulates these attributes to overwrite, or pollute, a JavaScript application object prototype of the base object by injecting other values. Properties on the Object.prototype are then inherited by all the JavaScript objects through the prototype chain. When that happens, this leads to either denial of service by triggering JavaScript exceptions, or it tampers with the application source code to force the code path that the attacker injects, thereby leading to remote code execution.

CVE-2022-22912
GHSA-4cpg-3vgw-4877
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -162,27 +162,6 @@
                 "node": ">=6.9.0"
             }
         },
-        "node_modules/@babel/core/node_modules/json5": {
-            "version": "2.2.0",
-            "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.0.tgz",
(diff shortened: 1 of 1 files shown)


veloce:
Vulnerability on dev dependency, no need to update.

PR #2246, opened 2022-11-23
Restore timeline by including `nb` query param

jas14:
See https://github.com/lichess-org/lila/pull/11879.


jas14:
cc @niklasf

niklasf:
Looks good. That'll do.

PR #2273, opened 2022-12-22
Support for new notifications

schlawg:


veloce:
There a lot of indentation changes. The only change really is: 
```
          default: {
             const url = action.notification.data['lichess.url'] as string
             if (url?.startsWith('https://lichess.org/')) {
               void openExternalBrowser(url)
             }
           }
``` 
right?

schlawg:
there is also probably an import openExternalBrowser at the top

veloce:
So the default behaviour is to open the browser within the app when one taps on a notif. I make sense for "streamer goes live" and maybe "study invite", but for "tournament starting soon" it's weird since the tournaments are in the app. 

Could we not open the tournament in the app instead?

veloce:
Isn't "tournamentStartingSoon" a new notification? the app never supported it.

schlawg:
Not sure.  That one predates my changes.

PR #2299, opened 2023-01-29
Add 12-character game link support

### LINKED ISSUE #2291: App does not capture 12-character full game IDs
# Current situation

Games are identified by 8 characters. However, these 8 characters can be [proceeded by another 4 characters](https://github.com/lichess-org/lila/blob/b707c9a277b52a61f63ab21b07784610fa7b2046/conf/routes#L304) indicating the player POV (essentially, white or black). URLs with these full game IDs are often shared from desktop or mobile web, since they're contained in the browser URL at the end of (or during) a game.

The app does not intercept this 12-character full game ID, so visiting such a link on mobile will direct you to whatever browser is installed instead of the native app experience.

# Suggestion

Intercept this app link and open the corresponding game, showing the corresponding perspective.

On desktop, when someone visits a URL with a POV that's not actually their own, they're [redirected](https://github.com/lichess-org/lila/blob/b707c9a277b52a61f63ab21b07784610fa7b2046/app/controllers/TheftPrevention.scala#L11) to the appropriate path, which is `/<gameId>/<color>` (or, in the case of white, just `/<gameId>`). Unfortunately, that is [not the case when accessing the game via the API](https://github.com/lichess-org/lila/blob/b707c9a277b52a61f63ab21b07784610fa7b2046/app/controllers/Round.scala#L60), but I imagine with a little server-side work, the API could return a reference to the game ID and the appropriate color perspective.

jas14:
Closes #2291. This should do the trick for Android, but we will need to update apple-app-site-associations on lila for iOS.

I tested on an Android emulator using the following command:
```
adb shell am start -W -a android.intent.action.VIEW -d "https://lichess.org/2360tJXP000z#50" org.lichess.mobileapp
```

and substituted various URIs, including:

* `https://lichess.org/2360tJXP000Z#50` (White's perspective at ply 50)
* `https://lichess.org/2360tJXP000Z` (White's perspective, no ply)
* `https://lichess.org/2360tJXP000z` (Black's perspective, no ply)

and some of the other link variations with just 8-character game IDs.


veloce:
I guess I should merge it when the apple-app-site-association is updated on Lila.

PR #2315, opened 2023-02-18
Adding more offline puzzles

### LINKED ISSUE #2256: Offline puzzle buffer should be ~200, not 50
**Is your feature request related to a problem? Please describe.**
I'm frustrated when I'm on a plane and I only have 50 offline puzzles to solve. I get through them, and then want to do more.

I also get into situations where the offline puzzles don't refresh for some reason, even when I was online, and that leads to me not being able to solve puzzles when I'm on a tube.

**Describe the solution you'd like**
Increase the number of offline puzzles to 200, form the current 50.

**Describe alternatives you've considered**
Have considered having a mode where you can replay the puzzles that you've gotten wrong in offline mode, which would increase the amount of time that you could spend offline and play puzzles. I've rejected that idea because that would be much more difficult to implement.

Have considered adding a setting feature where a user can toggle how many puzzles they'd like to save offline.

**Additional context**
This should just boil down to changing `puzzleBufferLen: 50,` to a higher number.

One drawback could be a small slowdown while getting the offline puzzles. I don't know exactly how high the payload is for 1 puzzle from a server, but I feel like it shouldn't be that much.

Rica320:
Closes #2256

If I understand correctly, replaying puzzles may not be a practical solution because puzzles likely contribute to one's overall rating.

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -393,7 +393,7 @@ export default {
   },
 
   training: {
-    puzzleBufferLen: 50,
+    puzzleBufferLen: 200,
     ratingDiffThreshold: 100
   },
 


ptrdom:
Could the size of the offline puzzle buffer be configurable through settings? Some constant number is bound to eventually not suit some users.

danik292 (approved):

PR #2362, opened 2023-05-21
Fix sorting and filtering studies

raylillywhite:
Previously, whenever you would use one of the filter or sort controls on the study page, it would reset the other one. That meant that it was impossible to view your own studies sorted by recently updated for example. This change preserves the filter when you change the sort and vice versa. Here's a video showing that the change works: 

https://github.com/lichess-org/lichobile/assets/504401/850eb6e4-23d7-4a57-8efb-64884981f268


jas14:
Looks like a good bug fix to me 😃

PR #2385, opened 2023-08-08
Convert the bug report template into a GitHub form

jas14:
It's hard to get people to reliably fill out enough information in the bug report to repro. Maybe this will help.

## Before
before-template

## After
after-form


veloce:
👍

PR #2469, opened 2024-06-04
Fixes after Timo tests

vovagorodok:


vovagorodok:
Sorry. By default github opens PR to lichobile:main

